PR #8, opened 2019-11-09
feat(disallow commands): add new parameter to secrets for injection

JordanSussman:
This adds the ability to disallow certain secrets from being injected if the specific step utilizes commands, you may want to do this to protect from someone doing the following malicious behavior:

```yaml
version: "1"

steps:
  - name: test
    image: alpine:latest
    secrets: [ super_secret ]
    commands:
      - env

secrets:
  - name: super_secret
    key:  super_secret
    engine: native
    type: repository
```

This particular logic presumes by default that you want secrets to be injected for steps with commands, so you would need to add the secret with `AllowCommands = false` to override that behavior.


JordanSussman:
Updated naming based on feedback from @jbrockopp.

PR #22, opened 2019-11-22
Add image column to steps and services tables

meinenec:
Changes on server ready

### DIFF
--- a/database/step_test.go
+++ b/database/step_test.go
@@ -20,6 +20,7 @@ func TestDatabase_Step_Nullify(t *testing.T) {
 		RepoID:       sql.NullInt64{Int64: 0, Valid: true},
 		Number:       sql.NullInt32{Int32: 0, Valid: true},
 		Name:         sql.NullString{String: "", Valid: true},
+		Image:        sql.NullString{String: "", Valid: true},
 		Stage:        sql.NullString{String: "", Valid: true},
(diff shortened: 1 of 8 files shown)


database/step_test.go
```diff
@@ -160,7 +166,7 @@ func TestDatabase_Step_Validate_NoRepoID(t *testing.T) {
 		ID:      sql.NullInt64{Int64: 1, Valid: true},
 		BuildID: sql.NullInt64{Int64: 1, Valid: true},
 		Number:  sql.NullInt32{Int32: 1, Valid: true},
-		Name:    sql.NullString{String: "foo", Valid: true},
```
jbrockopp: I believe this was removed in error 👍 

I recommend we keep the `Name` field here and add the `Image` field as well

database/step_test.go
```diff
@@ -176,7 +182,7 @@ func TestDatabase_Step_Validate_NoNumber(t *testing.T) {
 		ID:      sql.NullInt64{Int64: 1, Valid: true},
 		BuildID: sql.NullInt64{Int64: 1, Valid: true},
 		RepoID:  sql.NullInt64{Int64: 1, Valid: true},
-		Name:    sql.NullString{String: "foo", Valid: true},
```
jbrockopp: I believe this was removed in error 👍 

I recommend we keep the `Name` field here and add the `Image` field as well

PR #53, opened 2020-04-17
feat(library): add ToString helper function

jbrockopp:


library/string.go
```diff
@@ -0,0 +1,79 @@
...
+	case []interface{}:
+		return unmarshalSlice(v)
+	default:
+		return unmarshalMap(v)
+	}
+}
+
+// helper function to unmarshal a parameter in map format.
+func unmarshalMap(v interface{}) string {
+	yml, _ := yaml.Marshal(v)
```
kneal, line 59: Is there value in checking this error?

library/string.go
```diff
@@ -0,0 +1,79 @@
...
+	out, _ := json.YAMLToJSON(yml)
```
kneal, line 60: Is there value in checking this error?

library/string.go
```diff
@@ -0,0 +1,79 @@
...
+	out, _ := yaml.Marshal(v)
+
+	in := []string{}
+
+	err := yaml.Unmarshal(out, &in)
+	if err == nil {
+		return strings.Join(in, ",")
+	}
+
+	out, _ = json.YAMLToJSON(out)
```
kneal, line 76: Is there value in checking this error?

library/string.go
```diff
@@ -0,0 +1,79 @@
...
+func unmarshalMap(v interface{}) string {
+	yml, _ := yaml.Marshal(v)
+	out, _ := json.YAMLToJSON(yml)
+
+	return string(out)
+}
+
+// helper function to unmarshal a parameter in slice format.
+func unmarshalSlice(v interface{}) string {
+	out, _ := yaml.Marshal(v)
```
kneal, line 67: Is there value in checking this error?

library/string.go
jbrockopp: I don't think so 👍 The amount of cases where this will actually error out are relatively low.

This function is copied directly from the [go-vela/compiler](https://github.com/go-vela/compiler/blob/master/compiler/native/unmarshal.go#L57-L63) repo and we've been utilizing it for quite some time with no issues 🙂

library/string.go
jbrockopp: I don't think so 👍 The amount of cases where this will actually error out are relatively low.

This function is copied directly from the [go-vela/compiler](https://github.com/go-vela/compiler/blob/master/compiler/native/unmarshal.go#L57-L63) repo and we've been utilizing it for quite some time with no issues 🙂

library/string.go
jbrockopp: I don't think so 👍 The amount of cases where this will actually error out are relatively low.

This function is copied directly from the [go-vela/compiler](https://github.com/go-vela/compiler/blob/master/compiler/native/unmarshal.go#L65-L79) repo and we've been utilizing it for quite some time with no issues 🙂

library/string.go
jbrockopp: I don't think so 👍 The amount of cases where this will actually error out are relatively low.

This function is copied directly from the [go-vela/compiler](https://github.com/go-vela/compiler/blob/master/compiler/native/unmarshal.go#L65-L79) repo and we've been utilizing it for quite some time with no issues 🙂

library/string.go
kneal: Alright, Forgot about those in the compiler. I guess I still would learn towards adding error handling so in the rare chance it does happen we are safe from any potential panics that may occur from the functioning erroring. I'll leave it up to you though

jbrockopp: I think providing some context might help 👍

The `ToString()` function was intended to be a catch-all conversion for any Go type to a string. This function itself, gets used to convert all `environment` variables in a pipeline to a string because Docker and Kubernetes expect the `environment` variables to be a `map[string]string`.

Now let's circle back to the handling error portion:

Unfortunately, I'm not sure by adding error handling we are any safer from panics. This is mainly due to YAML itself supporting such a wide array of different key types (int, string, float etc.) where JSON only supports strings. Fortunately, both YAML libraries we're using here were intentionally written and designed to make it extremely difficult to panic. They do a fair amount of input sanitization to help prevent the possibility of this, and in the event they do panic, they're designed to recover.

To elaborate, let's take a look at the [`Marshal()`](https://github.com/go-yaml/yaml/blob/53403b58ad1b561927d19068c655246f2db79d48/yaml.go#L199-L207) function:

```go
func Marshal(in interface{}) (out []byte, err error) {
	defer handleErr(&err)
	e := newEncoder()
	defer e.destroy()
	e.marshalDoc("", reflect.ValueOf(in))
	e.finish()
	out = e.out
	return
}
```

The important line to call out here is the `defer handleErr(&err)` part. The [`handleErr()`](https://github.com/go-yaml/yaml/blob/v2.2.8/yaml.go#L244-L252) function looks like this:

```go
func handleErr(err *error) {
	if v := recover(); v != nil {
		if e, ok := v.(yamlError); ok {
			*err = e.err
		} else {
			panic(v)
		}
	}
}
```

From this function, the important part to call out is the [`recover()`](https://golang.org/pkg/builtin/#recover) function which is designed to recover from panics:

> The recover built-in function allows a program to manage behavior of a panicking goroutine. Executing a call to recover inside a deferred function (but not any function called by it) stops the panicking sequence by restoring normal execution and retrieves the error value passed to the call of panic. If recover is called outside the deferred function it will not stop a panicking sequence. In this case, or when the goroutine is not panicking, or if the argument supplied to panic was nil, recover returns nil. Thus the return value from recover reports whether the goroutine is panicking.

kneal: Context helps, still find it odd but I can understand why it's written the way it is now.

PR #66, opened 2020-05-11
enhancement(ruleset): add an execute func

kneal:
This change is to simplify and further test how `status` gets executed during a pipeline run.

Steps workflow:
https://github.com/go-vela/pkg-executor/blob/master/executor/linux/build.go#L386-L413

Stages workflow:
https://github.com/go-vela/pkg-executor/blob/master/executor/linux/stage.go#L117-L144

This charge also does some further testing to validate when the status is compared with other fields it can be properly processed.  

Some Fields **can't** be supported with ruleset execution because the data is not available to be processed. These fields are `comment` and `path`. Right now, we don't have a great way to propagate that data so if a `status` ruleset is used with those fields the user should expect the `status` may not be processed as expected.

### DIFF
--- a/pipeline/ruleset.go
+++ b/pipeline/ruleset.go
@@ -82,6 +82,27 @@ func (r *Ruleset) Match(from *RuleData) bool {
 	return false
 }
 
+// Execute returns true when the provided ruledata matches
+// the if rules and does not match any of the unless rules.
+// This function does not check When the provided if and Unless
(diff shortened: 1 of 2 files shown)


kneal:
Closing this PR in favor of different implementation

kneal:
Improved implementation:
https://github.com/go-vela/types/pull/69

PR #80, opened 2020-06-05
perf(library): clean output for String()

jbrockopp:
Currently, our `String()` functions for the resources don't provide a whole lot of value 😞 

For example, if you were to `fmt.Println()` on one of our library types, you'd see output similar to:

```go
{ID:0xc00002c808 Name:0xc0000103d0 Token:0xc0000103e0 Hash:0xc0000103f0 Favorites:0xc00000c320 Active:0xc00002c810 Admin:0xc00002c811}
```

**NOTE: This is the `go-vela/types/library.User` type being displayed.**

While it's helpful to see the fields inside the struct, unfortunately it only displays the memory address of the pointer rather than the actual underlying value.

After this PR is merged, the same output would now look like:

```go
{
  Active: true,
  Admin: false,
  Favorites: [github/octocat],
  ID: 1,
  Name: octocat,
  Token: superSecretToken,
}
```

I've attached a sample from Go playground so you can see the change:

https://play.golang.org/p/DR7mTEmCCc8


jbrockopp:
@wass3r normally, I would have used the `improvement` label but that [was removed from the conventional commit types](https://github.com/commitizen/conventional-commit-types/commit/958396a7224f6d6530b23998caffe82ddecc16e1) awhile back 😢 

[Based off the other commit types left](https://github.com/commitizen/conventional-commit-types/blob/master/index.json), `perf` is really the only one that fits.

I don't think `refactor` works here because it has a description of:

> A code change that neither fixes a bug nor adds a feature

I'd argue that although this change isn't fixing a bug, it's closer to "adding a feature" than not.

And `enhancement` isn't even a valid commit type so 🤷‍♂️

PR #81, opened 2020-06-11
feat: add plugin pattern for secrets look up

kneal:
This contains the necessary helper functions to compile and run secret plugins.

New syntax introduced:
```
secrets:
  - name: hello_world
    origin:
      name: vault
      image: target/secret-vault:latest
      secrets: [ vault_username, vault_password ]
      pull: true
      parameters:
        log_level: trace
        auth_method: ldap
        path: secret/hello_world
        keys: [ value ]
```

Related PRs:
 * https://github.com/go-vela/compiler/pull/76
* https://github.com/go-vela/pkg-executor/pull/41


pipeline/secret.go
```diff
@@ -12,11 +12,20 @@ type (
...
+		Key    string  `json:"key,omitempty"    yaml:"key,omitempty"`
+		Engine string  `json:"engine,omitempty" yaml:"engine,omitempty"`
+		Type   string  `json:"type,omitempty"   yaml:"type,omitempty"`
+		Origin *Origin `json:"origin,omitempty" yaml:"origin,omitempty"`
+	}
+
+	// Origin is the pipeline representation of a method
+	// for looking up secrets with a secret plugin.
+	Origin struct {
+		Image      string                 `yaml:"image,omitempty"`
```
jbrockopp, line 26: Could we add the `json` field tags here like we have for some other pipeline structs?

yaml/secret.go
```diff
@@ -24,6 +24,15 @@ type (
 		Key    string `yaml:"key,omitempty"`
 		Engine string `yaml:"engine,omitempty"`
 		Type   string `yaml:"type,omitempty"`
+		Origin Origin `yaml:"origin,omitempty"`
```
jbrockopp, line 27: Could we add the `json` field tags here like we have for some other pipeline structs?

yaml/secret.go
```diff
@@ -24,6 +24,15 @@ type (
...
+	}
+
+	// Origin is the yaml representation of a method
+	// for looking up secrets with a secret plugin.
+	Origin struct {
+		Image      string                 `yaml:"image,omitempty"`
```
jbrockopp, line 33: Could we add the `json` field tags here like we have for some other pipeline structs?

JordanSussman:
I'd argue that this should also handle the scenario where the plugin could return `map[string]interface{}` instead of just the raw value.

kneal:
Plugin returns should be handled within the plugin because the `map[string]interface{}` may be a secret provider-specific detail. This syntax is just to allow someone to run the plugin.

JordanSussman:
> Plugin returns should be handled within the plugin because the `map[string]interface{}` may be a secret provider-specific detail. This syntax is just to allow someone to run the plugin.

Suppose I'm a bit confused how this would work if [name](https://github.com/go-vela/types/pull/81/files#diff-3f206949ecb608c65f5d69f7db1ecfe0R15) is only key/value.

kneal:
So, the goal of the syntax is to look up a single value and store that as the `name:` which would then make it accessible in a step. So getting implementation-specific the order of operations might look like:

1. Iterate through secrets block looping up secrets
2. Use plugin to read the external secret
3. Set that secret value as ENV to be inspected by the executor
4. Executor extracts value and saves into a local secrets map


_Note: `3.` and `4.` allow us to keep our ACLs where the secret is never exposed to a container that isn't asking for it._

Then once the name is saved in the map the executor has the ability to sense when to inject the secret into a specific step. Today, that map is a `map[string]library.Secret` I don't think that will change but I haven't gotten to that part of the implementation for the feature yet. 

Changing `name:` to be a `map[string]interface{}` would be a significantly larger change that would affect how we store/use that secret in general.

JordanSussman:
> Changing name: to be a map[string]interface{} would be a significantly larger change that would affect how we store/use that secret in general.

IMO this plugin architecture has significantly less value if it does not account for scenarios where the external system (vault) will return `map[string]interface{}`. However, I can definitely see the point that implementing that change is significant compared to continuing to extend the current behavior.

kneal:
This feature is still early on so I have a few ideas on enable the secret provider to store objects (`interface`) and not conform to Vela's key/value approach to storage. The main idea is if a user doesn't provide the `name:` during the look up process. The plugin needs to be responsible for sourcing all desired secret values with a `SECRET_` prefix in the environment. This would then allow the platform (Vela) to extract only the data it needs and allow them to be used in steps.

The user would need to more than likely need to use the `source:` and `target:` syntax in the step secret block to turn their `secret_name_of_secret` to `name_of_secret` when they want that secret to be used in a step.

JordanSussman:
> This feature is still early on so I have a few ideas on how to enable the secret providers to store objects (`interface`) and not conform to Vela's key/value approach to storage. The main idea is if a user doesn't provide the `name:` during the look up process. The plugin needs to be responsible for sourcing all desired secret values with a `SECRET_` prefix in the environment. This would then allow the platform (Vela) to extract only the data it needs and allow them to be used in steps.
> 
> The user would need to more than likely need to use the `source:` and `target:` syntax in the step secret block to turn their `secret_name_of_secret` to `name_of_secret` when they want that secret to be used in a step.

You may also contemplate adding the additional parameter of `prefix` that could be utilized instead of `name`. The aforementioned `prefix` could mutate the variables that are returned by the plugin to be `<PREFIX>_<NAME RETURNED>=<VALUE RETURNED>`.

pipeline/container.go
```diff
@@ -82,48 +83,87 @@ func (c *ContainerSlice) Purge(r *RuleData) *ContainerSlice {
...
+		return container
 	// unrecognized driver
 	default:
 		// log here?
 		return nil
 	}
 }
 
+// Empty returns true if the provided ruletypes are empty.
+func (c *Container) Empty() bool {
```
kneal, line 139: The function is used to check if the container is empty. The function is typically used when looping through secret arrays to identify secret plugin lookups.

pipeline/secret.go
```diff
@@ -38,3 +47,158 @@ type (
...
+	// org in key does not equal the name of the team.
+	ErrInvalidShared = errors.New("invalid team in key")
+	// ErrInvalidPath defines the error type when the
+	// path provided for a type (org, repo, shared) is invalid.
+	ErrInvalidPath = errors.New("invalid secret path")
+)
+
+// Purge removes the secrets that have a ruleset
+// that do not match the provided ruledata.
+func (s *SecretSlice) Purge(r *RuleData) *SecretSlice {
```
kneal, line 71: Used to remove the secret containers that do not match the build event

pipeline/secret.go
```diff
@@ -38,3 +47,158 @@ type (
...
+			*secrets = append(*secrets, secret)
+		}
+	}
+
+	return secrets
+}
+
+// ValidOrg returns true when the secret is valid for a given
+// organization.
+func (s *Secret) ValidOrg(org string) error {
```
kneal, line 102: Functionality was moved from pkg-executor to types to allow cleaner validation on secrets

pipeline/secret.go
```diff
@@ -38,3 +47,158 @@ type (
...
+	if !strings.EqualFold(parts[0], org) {
+		return fmt.Errorf("%s: %s ", ErrInvalidOrg, org)
+	}
+
+	return nil
+}
+
+// ValidRepo returns an error when the secret is valid for a given
+// organization and repository.
+func (s *Secret) ValidRepo(org, repo string) error {
```
kneal, line 134: Functionality was moved from pkg-executor to types to allow cleaner validation on secrets

pipeline/secret.go
```diff
@@ -38,3 +47,158 @@ type (
...
+	if !strings.EqualFold(s.Name, s.Key) {
+		return fmt.Errorf("%s: %s ", ErrInvalidPath, path)
+	}
+
+	return nil
+}
+
+// ValidShared returns true when the secret is valid for a given
+// organization and team.
+func (s *Secret) ValidShared(org string) error {
```
kneal, line 176: Functionality was moved from pkg-executor to types to allow cleaner validation on secrets

yaml/secret.go
```diff
@@ -82,6 +96,33 @@ func (s *SecretSlice) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		o.Parameters == nil &&
+		len(o.Secrets) == 0 &&
+		o.Pull == false {
+		return true
+	}
+
+	return false
+}
+
+// Empty returns true if the provided origin is empty.
```
jbrockopp, line 114: I think this comment needs to be updated 👍

yaml/secret.go
```diff
@@ -82,6 +96,33 @@ func (s *SecretSlice) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+// Empty returns true if the provided origin is empty.
+func (o *Origin) Empty() bool {
+	// return true if every ruletype is empty
```
jbrockopp, line 101: I think this comment needs to be updated 👍

pipeline/secret.go
```diff
@@ -38,3 +47,158 @@ type (
...
+	// check if name equals key for implicit definition
+	if !strings.EqualFold(s.Name, s.Key) {
+		return fmt.Errorf("%s: %s ", ErrInvalidPath, path)
+	}
+
+	return nil
+}
+
+// ValidShared returns true when the secret is valid for a given
+// organization and team.
```
jbrockopp, line 175: I think this comment needs to be updated 👍

pipeline/secret.go
```diff
@@ -38,3 +47,158 @@ type (
...
+	// check if the org provided matches what we expect
+	if !strings.EqualFold(parts[0], org) {
+		return fmt.Errorf("%s: %s ", ErrInvalidOrg, org)
+	}
+
+	return nil
+}
+
+// ValidRepo returns an error when the secret is valid for a given
+// organization and repository.
```
jbrockopp, line 133: I think this comment needs to be updated 👍

pipeline/secret.go
```diff
@@ -38,3 +47,158 @@ type (
...
+			// append the secret to the new slice of secrets
+			*secrets = append(*secrets, secret)
+		}
+	}
+
+	return secrets
+}
+
+// ValidOrg returns true when the secret is valid for a given
+// organization.
```
jbrockopp, line 101: I think this comment needs to be updated 👍

pipeline/container.go
```diff
@@ -82,48 +83,87 @@ func (c *ContainerSlice) Purge(r *RuleData) *ContainerSlice {
...
+	// return true if every ruletype is empty
```
jbrockopp, line 140: I think this comment needs to be updated 👍

pipeline/container.go
```diff
@@ -82,48 +83,87 @@ func (c *ContainerSlice) Purge(r *RuleData) *ContainerSlice {
...
+}
+
+// Sanitize cleans the fields for every step in the pipeline so they
+// can be safely executed on the worker. The fields are sanitized
+// based off of the provided runtime driver which is setup on every
+// worker. Currently, this function supports the following runtimes:
+//
+//   * Docker
+//   * Kubernetes
+func (c *Container) Sanitize(driver string) *Container {
```
jbrockopp, line 105: Good idea! I like this better 👍

kneal: In the future, I think we could do this to the `Purge` function as well to break the code up into smaller functions.

kneal:
@wass3r 

> since we're already running reflect.DeepEqual, couldn't we run that on the whole of Container and compare to Container{}?

Yes, we could but we also have the pattern of checking for empty by checking each value:

* [Ruleset empty](https://github.com/go-vela/types/blob/master/pipeline/ruleset.go#L102-L118)
* [Worker empty](https://github.com/go-vela/types/blob/master/pipeline/worker.go#L15-L25)

So we have a couple of different patterns in place. We probably want to be consistent across the board in those helper functions.

wass3r:
@kneal, thanks. it's not necessarily stopping me from approving it, just wanted more context. it's a little more burdensome and does allow for missed checks when the respective struct changes. could be a separate story, if we wanted to change that up, though.

kneal:
Yeah, I'm not opposed to that pattern since we do it in unit testing. Does Go have any rules against/for using reflection in that way?

wass3r:
@kneal - i haven't checked too closely and don't have data, but i think using `reflect.DeepEqual` probably performs worse than checking each value, if you're asking from a performance perspective. Since we're dealing with non-comparables i think the options are limited to reflect or property checks. we can go all in on property checking, too. then, i'd propose changing the current reflection check on Ruleset (now, or later).

i like the reflection way more from a maintenance perspective, so probably have to weight trade-off between (potential?) performance hit and maintenance.

to answer the question though, i'm not aware of any official rules.

jbrockopp:
@kneal @wass3r 

I'm good either way on how we want to define checking if a resource is empty 👍 

As far as Go having rules or guidelines on using the reflection library in this way, I don't believe there is anything explicitly stated.

I think the only concern for using `reflect.DeepEqual` would be the potential hit by performance.

To be clear, I'm not saying that using it here would offer less performance.

But, it could take a hit in performance because that function accepts `interface{}` as the arguments. This means that it will then have extra logic under the hood to determine the type of the arguments being provided before it can compare them which can be costly.

All that being said, I'm good either way because I imagine any performance hit (if any existed) would be very, very minimal 👍

PR #85, opened 2020-07-20
feat(build): track head and base ref for pr

JordanSussman:
## summary

Adds the `VELA_PULL_REQUEST_SOURCE` and `VELA_PULL_REQUEST_TARGET` variables that match the base and head refs from [go-github](https://github.com/google/go-github/blob/master/github/pulls.go#L72-L73). 

Starts the work to complete https://github.com/go-vela/community/issues/28.

## value

Allows plugins/users to determine both branches involved in a given pull request.

Note: The failing reviewdog action is unrelated to my changes.

### DIFF
--- a/library/build.go
+++ b/library/build.go
@@ -39,6 +39,7 @@ type Build struct {
 	Branch       *string `json:"branch,omitempty"`
 	Ref          *string `json:"ref,omitempty"`
 	BaseRef      *string `json:"base_ref,omitempty"`
+	HeadRef      *string `json:"head_ref,omitempty"`
 	Host         *string `json:"host,omitempty"`
 	Runtime      *string `json:"runtime,omitempty"`
 	Distribution *string `json:"distribution,omitempty"`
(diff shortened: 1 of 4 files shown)


library/build.go
```diff
@@ -39,6 +39,7 @@ type Build struct {
 	Branch       *string `json:"branch,omitempty"`
 	Ref          *string `json:"ref,omitempty"`
 	BaseRef      *string `json:"base_ref,omitempty"`
+	HeadRef      *string `json:"head_ref,omitempty"`
```
jbrockopp, line 42: I think if we want to add a new field to this table, we'll also need to do so for the `database` type as well 👍 

Were you planning to add that in a separate PR?

You can find this other type in the `go-vela/types/database` package 😃 

https://github.com/go-vela/types/blob/6af7ad668990cadb758a68e5a9a532a1eb3237ca/database/build.go#L24-L53

jbrockopp:
@JordanSussman thanks for your contribution! 🎉

PR #86, opened 2020-07-22
feat(skip): add ability to skip hook

wass3r:
pre-work for https://github.com/go-vela/community/issues/66 and https://github.com/go-vela/community/issues/70

will be functional once associated, incoming PR to server is merged (https://github.com/go-vela/server/pull/169)

- adds method on Webhook struct called `ShouldSkip`
- adds helper function to check string for skip directive
- adds tests to check `ShouldSkip` functionality

### DIFF
--- a/webhook.go
+++ b/webhook.go
@@ -4,7 +4,17 @@
 
 package types
 
-import "github.com/go-vela/types/library"
+import (
+	"strings"
+
+	"github.com/go-vela/types/constants"
+	"github.com/go-vela/types/library"
+)
+
+var (
+	skipDeleteEventMsg = "tag/branch delete event"
+	skipDirectiveMsg   = "skip ci directive found in commit title/message"
+)
 
 // Webhook defines a struct that is used to return
 // the required data when processing webhook event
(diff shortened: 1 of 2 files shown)


webhook.go
```diff
@@ -16,3 +26,45 @@ type Webhook struct {
...
+// ShouldSkip uses the build information
+// associated with the given hook to determine
+// whether the hook should be skipped
+func (w *Webhook) ShouldSkip() (bool, string) {
+	// push event
+	if strings.EqualFold(constants.EventPush, w.Build.GetEvent()) {
+		// the head commit will return null in the hook
+		// payload from the scm when the event is
+		// associated with a branch/tag delete
+		if len(w.Build.GetCommit()) == 0 {
```
wass3r, line 39: trying to use what we have available on the hook here. the original payload has some other attributes one could consider, such as the `deleted` field, or you could cue off of the `after` field which will be a blank commit sha consisting of all zeroes (see example payload here: https://developer.github.com/webhooks/event-payloads/#webhook-payload-example-32)

wass3r:
@jbrockopp thanks for taking a look and providing the other references.

i'm not sure either. that is why i had a lengthy note there :)

i guess, i'd prefer that the `Webhook` struct (or the `Build` struct) has all the info needed to determine whether it should be skipped, not much different than a `Validate` method on a struct. then, we can have it all in one place. however, that's probably not feasible with the current abstraction. i was thinking of lifting the `before` and `after` shas from the payload and attaching it to the `Build` struct (new fields) because those values seem reasonable to expect across the board from vcs/scm systems. then we could cue off of that for this scenario. unfortunately, that approach would not accomodate what we're doing for PRs that you highlighted (if the goal was to shift skip logic here).

(side note: it would be nice to surface specific reasons for why a build didn't occur vs the generic `no build to process`)

you didn't mention the "[skip ci]" parts in your comments? any thoughts there?

kneal:
The thing I like most about this implementation is it allows us to potentially surface those skip messages in the webhooks logs or an activity page within the application. I do think it may be valuable to users to see that Vela received the payload processed something but the commit/event wasn’t chosen to kick off a build.

jbrockopp:
@wass3r sure I think that's fair and I don't want to be the stick in the mud here 😄 

If I'm not mistaken, it sounds like the preference would be to push the logic for skipping both `push` and `pull_request` builds (based off certain criteria) here. I'm not opposed to that strategy either, so my question would be, do we think that should be a separate story?

i.e. Go with the PRs you have in this repo and in `go-vela/server` (MVP right? 😃 ) and then a separate story would be added to address making the process of skipping builds consistent? Meaning move the rest of the PR logic to this function at a later time.

If you feel compelled enough to do it now, then I'd say go for it. But I wasn't trying to make your life harder when I asked about the other logic we have for skipping PRs 😅 

> you didn't mention the "[skip ci]" parts in your comments? any thoughts there?

You are right and that's because I like what you have for the logic. I'm just not entirely sure where that logic belongs so I thought I'd ask about the `deleted` attribute in the webhook payload and go from there 👍

wass3r:
@jbrockopp @kneal thanks again for the comments.

yea, i think a separate story is good for the pull request aspect. probably a couple different ways to tackle and i'm not sure which way to go on that at the moment.

kind of open question still then, at least in my mind. are we good with using `GetCommit` which will be empty due to `head_commit: null` (at least for Github's payload) as a way to determine delete event? i couldn't get that to be the case for any other condition.

jbrockopp:
@wass3r that works for me 👍

wass3r:
story created: https://github.com/go-vela/community/issues/78

PR #87, opened 2020-07-27
feat(token): add refresh token info

wass3r:
the additions here can be summarized as:
1. add a constant for the refresh token cookie name
2. adds a RefreshToken field to keep track of in the database
3. remove _now_ unused code related to old login method
4. adds a WebOauthCallbackPath field to the metadata to allow for customizing the callback destination in case other UIs have different structure


kneal (approved):
LGTM 🐬

jbrockopp (approved):
LGTM

PR #105, opened 2020-08-22
fix(schema): stages don't require name field

wass3r:
name is set to key if not supplied as evidenced by: https://github.com/go-vela/types/blob/master/yaml/stage.go#L78

### DIFF
--- a/yaml/stage.go
+++ b/yaml/stage.go
@@ -22,7 +22,7 @@ type (
 	// of a stage in a pipeline.
 	// nolint:lll // jsonschema will cause long lines
 	Stage struct {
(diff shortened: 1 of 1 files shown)


kneal:
@wass3r Yeah, you're right looks like it gets implicitly set
https://github.com/go-vela/types/blob/master/yaml/stage.go#L76-L79

wass3r:
Cool, yea.. we require it internally, but from a user's perspective it's optional. sorry for confusion.

kneal:
No problem! I just remembered seeing the `name:` be used in areas of the compiler and didn’t think to look if we implicitly set it when it’s missing.

PR #106, opened 2020-09-08
feat: add worker type

meinenec:
struct representing a worker stored in the database.

this is not the same as the struct contained in the [worker](https://github.com/go-vela/worker/blob/master/cmd/vela-worker/worker.go#L61-L67) codebase.

### DIFF
--- a/database/worker.go
+++ b/database/worker.go
@@ -8,6 +8,7 @@ import (
 	"database/sql"
 
 	"github.com/go-vela/types/library"
+	"github.com/lib/pq"
 )
 
 // Worker is the database representation of a worker.
@@ -15,6 +16,7 @@ type Worker struct {
 	ID            sql.NullInt64  `sql:"id"`
 	Hostname      sql.NullString `sql:"hostname"`
 	Address       sql.NullString `sql:"address"`
+	Routes        pq.StringArray `sql:"routes"`
 	Active        sql.NullBool   `sql:"active"`
 	LastCheckedIn sql.NullTime   `sql:"last_checked_in"`
 }
@@ -40,7 +42,7 @@ func (w *Worker) Nullify() *Worker {
 		w.Hostname.Valid = false
 	}
 
-	// check if the URL field should be false
+	// check if the Address field should be false
 	if len(w.Address.String) == 0 {
 		w.Hostname.Valid = false
 	}
@@ -56,6 +58,7 @@ func (w *Worker) ToLibrary() *library.Worker {
 	worker.SetID(w.ID.Int64)
 	worker.SetHostname(w.Hostname.String)
 	worker.SetAddress(w.Address.String)
+	worker.SetRoutes(w.Routes)
 	worker.SetActive(w.Active.Bool)
 	worker.SetLastCheckedIn(w.LastCheckedIn.Time)
 	return worker
@@ -68,6 +71,7 @@ func WorkerFromLibrary(w *library.Worker) *Worker {
 		ID:            sql.NullInt64{Int64: w.GetID(), Valid: true},
 		Hostname:      sql.NullString{String: w.GetHostname(), Valid: true},
 		Address:       sql.NullString{String: w.GetAddress(), Valid: true},
+		Routes:        w.GetRoutes(),
 		Active:        sql.NullBool{Bool: w.GetActive(), Valid: true},
 		LastCheckedIn: sql.NullTime{Time: w.GetLastCheckedIn(), Valid: true},
 	}
--- a/database/worker.go
+++ b/database/worker.go
@@ -15,7 +15,7 @@ type Worker struct {
 	ID            sql.NullInt64  `sql:"id"`
 	Hostname      sql.NullString `sql:"hostname"`
 	Address       sql.NullString `sql:"address"`
-	Online        sql.NullBool   `sql:"online"`
+	Active        sql.NullBool   `sql:"active"`
 	LastCheckedIn sql.NullTime   `sql:"last_checked_in"`
 }
 
@@ -56,7 +56,7 @@ func (w *Worker) ToLibrary() *library.Worker {
 	worker.SetID(w.ID.Int64)
 	worker.SetHostname(w.Hostname.String)
 	worker.SetAddress(w.Address.String)
-	worker.SetOnline(w.Online.Bool)
+	worker.SetActive(w.Active.Bool)
 	worker.SetLastCheckedIn(w.LastCheckedIn.Time)
 	return worker
 }
@@ -68,7 +68,7 @@ func WorkerFromLibrary(w *library.Worker) *Worker {
 		ID:            sql.NullInt64{Int64: w.GetID(), Valid: true},
 		Hostname:      sql.NullString{String: w.GetHostname(), Valid: true},
 		Address:       sql.NullString{String: w.GetAddress(), Valid: true},
-		Online:        sql.NullBool{Bool: w.GetOnline(), Valid: true},
+		Active:        sql.NullBool{Bool: w.GetActive(), Valid: true},
 		LastCheckedIn: sql.NullTime{Time: w.GetLastCheckedIn(), Valid: true},
 	}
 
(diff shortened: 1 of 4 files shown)


database/worker.go
```diff
@@ -0,0 +1,76 @@
...
+	"database/sql"
+
+	"github.com/go-vela/types/library"
+)
+
+// Worker is the database representation of a worker.
+type Worker struct {
+	ID            sql.NullInt64  `sql:"id"`
+	Hostname      sql.NullString `sql:"hostname"`
+	Path          sql.NullString `sql:"path"`
```
kneal, line 17: Did you have any other ideas than `path` to hold the address of a worker? `path` doesn't seem very intuitive to me personally.

meinenec: i'm open to suggestions. was originally url, but i wasn't thrilled about either choice.

kneal: I would prefer url over path. The only ideas I have would be like address, URL, ingress, location.

database/worker.go
```diff
@@ -0,0 +1,76 @@
...
+package database
+
+import (
+	"database/sql"
+
+	"github.com/go-vela/types/library"
+)
+
+// Worker is the database representation of a worker.
+type Worker struct {
```
JordanSussman, line 14: Should this also contain information about the route(s) the worker is a part of? Thinking of situations where you potentially want to do maintenance on a subset of the workers.

database/worker.go
jbrockopp: I agree that I'm not a big fan of `path` since it could cause confusion.

I'm thinking specifically in technology, `path` often refers to an actual path to a file or directory on a host.

`URL` or `Address` would work for me 👍

JordanSussman: My vote would be for address.

database/worker.go
```diff
@@ -0,0 +1,76 @@
...
+	ID            sql.NullInt64  `sql:"id"`
+	Hostname      sql.NullString `sql:"hostname"`
+	Path          sql.NullString `sql:"path"`
+	Online        sql.NullBool   `sql:"online"`
```
jbrockopp, line 18: Could we change this to `Active` instead of `Online`?

`Active` fits better with some of the field names we already have in other types 👍 

https://github.com/go-vela/types/blob/3e7f85be5d18e6fe39e992946a2b7eaa903800d7/database/repo.go#L55

https://github.com/go-vela/types/blob/3e7f85be5d18e6fe39e992946a2b7eaa903800d7/database/user.go#L50

database/worker.go
```diff
@@ -0,0 +1,76 @@
...
+	worker := &Worker{
+		ID:            sql.NullInt64{Int64: w.GetID(), Valid: true},
+		Hostname:      sql.NullString{String: w.GetHostname(), Valid: true},
+		Path:          sql.NullString{String: w.GetPath(), Valid: true},
+		Online:        sql.NullBool{Bool: w.GetOnline(), Valid: true},
+		LastCheckedIn: sql.NullTime{Time: w.GetLastCheckedIn(), Valid: true},
+	}
+
+	return worker.Nullify()
+}
```
jbrockopp, line 76: I think you'll want to add a `Validate()` function similar to what we have for the other types 👍 

The purpose of these functions is to quickly validate the required fields for the struct have content before creating/updating in the database.

Some examples:

https://github.com/go-vela/types/blob/3e7f85be5d18e6fe39e992946a2b7eaa903800d7/database/build.go#L263-L277

https://github.com/go-vela/types/blob/3e7f85be5d18e6fe39e992946a2b7eaa903800d7/database/hook.go#L139-L158

https://github.com/go-vela/types/blob/3e7f85be5d18e6fe39e992946a2b7eaa903800d7/database/repo.go#L160-L194

jbrockopp (changes requested):

database/worker.go
meinenec: I'm okay with address, I'll change to that

database/worker.go
meinenec: That's doable. Makes sense too!

database/worker.go
meinenec: Sure, I think that would be more useful since I can imagine a situation in the future where a worker might be "online" but disabled.

database/worker.go
meinenec: I went back and forth on this so thanks for mentioning it.

I had looked at those examples and noticed that only required fields are validated. In this case, what fields would be required? My initial thought might be that `Hostname` and `Path` (soon to be `Address`) are the only two that might be considered required, but I'm not sure of a scenario where all fields wouldn't be set and couldn't be validated.

jbrockopp: That's a great question that I'm also uncertain what the answer should be 😛 

I think what you have initially would be good (require `Hostname` and `Address`) and we can always expand upon it later!

I too often find myself going through the mental gymnastics of trying to determine what should and shouldn't be considered a required field. The way I try to think about the required fields in a struct is more of a functionality perspective and what is actually necessary to perform actions within Vela.

I know that's a bit vague so to be more helpful, let's apply context from this specific situation to better illustrate 😃 

From a functionality perspective, I think to myself:

> What fields in the database table will actually be used in the code?

To answer, I imagine the `Address` is absolutely necessary because that's the value we'll use to send API requests from the server to the worker 👍 

Since the `Address` itself will also contain the `Hostname`, I think it's reasonable to also say that would be required.

As we start to evaluate the other fields, I'm unsure we can apply the same "required" label to it but that doesn't mean those fields aren't important or have a purpose.

For example, would `LastCheckedIn` or `Active` be things that are actually used in the code?

The answer is probably not although they are still useful and serve a purpose, especially from an admin perspective 👍 

And to your point, in a perfect world, **all fields** should be set and could be validated, but for now, I'd recommend only validating what is absolutely required 😄

meinenec: thanks for the explanation! I could see `Active` being used when the server goes to connect to a worker. It might check to see that the worker is marked as active before attempting to connect and waiting for a timeout, but other than that I don't see those being used currently.

I'll add the Address and Hostname fields for now. Like you mentioned we can always add more so I'll get  the function in place to make that easier.

jbrockopp: That's a fair point 👍 

We do use the `Active` field in other parts of the codebase to make sure it is `true` before proceeding with more actions.

But sounds like you got it figured out and know where to go from here! 🎉 

If you have more questions let me know otherwise I'll be on the lookout for more changes 👀

database/worker.go
```diff
@@ -0,0 +1,96 @@
...
+	worker.SetLastCheckedIn(w.LastCheckedIn.Time)
+	return worker
+}
+
+// Validate verifies the necessary fields for
+// the Repo type are populated correctly.
+func (w *Worker) Validate() error {
+	// verify the Host field is populated
+	if len(w.Hostname.String) == 0 {
+		return ErrEmptyRepoHash
```
jbrockopp, line 72: I think we might want to change this error message 👍 

Typically we create unique error messages per field per struct.

Currently, it looks like we're referencing this error:

https://github.com/go-vela/types/blob/3e7f85be5d18e6fe39e992946a2b7eaa903800d7/database/repo.go#L19-L21

database/worker.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// verify the Address field is populated
+	if len(w.Address.String) == 0 {
+		return ErrEmptyRepoOrg
```
jbrockopp, line 77: I think we might want to change this error message 👍 

Typically we create unique error messages per field per struct.

Currently, it looks like we're referencing this error:

https://github.com/go-vela/types/blob/3e7f85be5d18e6fe39e992946a2b7eaa903800d7/database/repo.go#L27-L29

jbrockopp (changes requested):

database/worker.go
meinenec: Thanks for catching these.

jbrockopp (approved):
LGTM

wass3r (approved):

PR #110, opened 2020-09-28
fix: issue with accidental secret exposure via wrong syntax

tvangtarget:
When using the method to inject a secret with alternative name, ie.

```
secrets:
  - source: secret
    target: alt_secret
```

it's possible to expose the secret by doing the following:

```
secrets:
  - source: secret
  - target: alt_secret
```

### DIFF
--- a/yaml/secret.go
+++ b/yaml/secret.go
@@ -207,7 +207,6 @@ func (s *StepSecretSlice) UnmarshalYAML(unmarshal func(interface{}) error) error
 	// attempt to unmarshal as a step secret slice type
 	err = unmarshal(secrets)
 	if err == nil {
-
 		// check for secret source and target
 		for _, secret := range *secrets {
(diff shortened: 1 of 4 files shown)


wass3r (dismissed):

yaml/secret.go
```diff
@@ -206,6 +207,14 @@ func (s *StepSecretSlice) UnmarshalYAML(unmarshal func(interface{}) error) error
 	// attempt to unmarshal as a step secret slice type
 	err = unmarshal(secrets)
 	if err == nil {
+
```
wass3r, line 210: lgtm once this lint error is fixed 👍

tvangtarget: fixed 👍

jbrockopp (approved):
LGTM

wass3r (approved):

PR #112, opened 2020-10-15
feat: add version pkg for Vela

jbrockopp:
Implementing a custom `Version` type for Vela that can be referenced throughout the different repos.

When rendered as JSON the output looks like:

```json
{
  "canonical": "v1.2.3",
  "major": 1,
  "minor": 2,
  "patch": 3,
  "metadata": {
    "architecture": "amd64",
    "build_date": "1970-1-1T00:00:00Z",
    "compiler": "gc",
    "git_commit": "abcdef123456789",
    "go_version": "1.15.0",
    "operating_system": "linux"
  }
}
```

Beyond the `String()` functions, this package also supports two different functions:

* `Meta()` - returns only the metadata surrounding the version of the application
* `Semantic()` - returns only the canonical semantic version of the application

### DIFF
--- a/version/metadata.go
+++ b/version/metadata.go
@@ -1,0 +1,45 @@
+// Copyright (c) 2020 Target Brands, Inc. All rights reserved.
+//
+// Use of this source code is governed by the LICENSE file in this repository.
+
+package version
+
+import "fmt"
+
+const metaFormat = `{
+  Architecture: %s,
+  BuildDate: %s,
+  Compiler: %s,
+  GitCommit: %s,
+  GoVersion: %s,
+  OperatingSystem: %s,
+}`
+
+// Metadata represents extra information surrounding the application version.
+type Metadata struct {
+	// Architecture represents the architecture information for the application.
+	Architecture string `json:"architecture,omitempty"`
+	// BuildDate represents the build date information for the application.
+	BuildDate string `json:"build_date,omitempty"`
+	// Compiler represents the compiler information for the application.
+	Compiler string `json:"compiler,omitempty"`
+	// GitCommit represents the git commit information for the application.
+	GitCommit string `json:"git_commit,omitempty"`
+	// GoVersion represents the golang version information for the application.
+	GoVersion string `json:"go_version,omitempty"`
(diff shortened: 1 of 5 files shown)


version/metadata.go
```diff
@@ -0,0 +1,45 @@
...
+import "fmt"
+
+const metaFormat = `{
+  Architecture: %s,
+  BuildDate: %s,
+  Compiler: %s,
+  GitCommit: %s,
+  GoVersion: %s,
+  OperatingSystem: %s,
+}`
```
jbrockopp, line 16: This is the format if we were to output the `Metadata` from the struct.

version/version.go
```diff
@@ -0,0 +1,71 @@
...
+  PreRelease: %s,
+  Metadata: {
+    Architecture: %s,
+    BuildDate: %s,
+    Compiler: %s,
+    GitCommit: %s,
+    GoVersion: %s,
+    OperatingSystem: %s,
+  }
+}`
```
jbrockopp, line 25: This is the format if we were to output the entire `Version` struct.

version/version.go
```diff
@@ -0,0 +1,71 @@
...
+	// PreRelease represents unstable changes that might not be compatible.
+	PreRelease string `json:"pre_release,omitempty"`
+	// Metadata represents extra information surrounding the application version.
+	Metadata Metadata `json:"metadata,omitempty"`
+}
+
+// Meta implements a formatted string containing only metadata for the Version type.
+func (v *Version) Meta() string {
+	return v.Metadata.String()
+}
```
jbrockopp, line 48: Here is the function used to return only the `Metadata` from the version.

version/version.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Semantic implements a formatted string containing a formal semantic version for the Version type.
+func (v *Version) Semantic() string {
+	return v.Canonical
+}
```
jbrockopp, line 53: Here is the function used to return only the canonical reference of the semantic version.

wass3r (approved):

kneal (approved):
LGTM 🐬

PR #113, opened 2020-10-21
feat: add a format type for starlark

kneal:
Co-authored-by: @JordanSussman
Co-authored-by: @antoniobeyah
Co-authored-by: @meinenec


yaml/template.go
```diff
@@ -19,6 +19,7 @@ type (
 	Template struct {
 		Name   string `yaml:"name,omitempty"   json:"name,omitempty"  jsonschema:"required,minLength=1,description=Unique identifier for the template.\nReference: https://go-vela.github.io/docs/concepts/pipeline/templates/"`
 		Source string `yaml:"source,omitempty" json:"source,omitempty" jsonschema:"required,minLength=1,description=Path to template in remote system.\nReference: https://go-vela.github.io/docs/concepts/pipeline/templates/source/"`
+		Format string `yaml:"format,omitempty"   json:"format,omitempty" jsonschema:"minLength=1,description=language used within the template file \nReference: https://go-vela.github.io/docs/concepts/pipeline/templates/format/,example=go,starlark"`
```
wass3r, line 22: i think we might want an enum definition here if there is a finite two options for this right now. see https://github.com/go-vela/types/blob/v0.6.0/yaml/step.go#L32 for an example.

wass3r (changes requested):

yaml/template.go
kneal: Right now, we have it set up as mainly two options but one has different permutations:

`format: starlark` will ensure the template gets expanded with that code path.

Go however has several options since we wanted to ensure backward compatibility for all existing templates. So, the following are valid to use a Go template:

```
format: golang
format: go
format: ""
```

wass3r (approved):

JordanSussman (approved):

yaml/template.go
```diff
@@ -19,6 +19,7 @@ type (
 	Template struct {
 		Name   string `yaml:"name,omitempty"   json:"name,omitempty"  jsonschema:"required,minLength=1,description=Unique identifier for the template.\nReference: https://go-vela.github.io/docs/concepts/pipeline/templates/"`
 		Source string `yaml:"source,omitempty" json:"source,omitempty" jsonschema:"required,minLength=1,description=Path to template in remote system.\nReference: https://go-vela.github.io/docs/concepts/pipeline/templates/source/"`
+		Format string `yaml:"format,omitempty"   json:"format,omitempty" jsonschema:"enum=starlark,enum=golang,default=go,minLength=1,description=language used within the template file \nReference: https://go-vela.github.io/docs/concepts/pipeline/templates/format/"`
```
jbrockopp, line 22: Not gonna lie.. my OCD is screaming at me right now for this 🤣 

```suggestion
		Format string `yaml:"format,omitempty"  json:"format,omitempty" jsonschema:"enum=starlark,enum=golang,default=go,minLength=1,description=language used within the template file \nReference: https://go-vela.github.io/docs/concepts/pipeline/templates/format/"`
```

Ahhh... there we go!

jbrockopp (approved):
LGTM! Ignore my OCD

PR #114, opened 2020-10-30
chore: silencing the linter

wass3r:
Attempting to address all existing linter issues. Any subsequent linter issues will be net new and should be handled before merge. Might even consider making linter a required check?

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -40,7 +40,7 @@ linters-settings:
     settings:
       mnd:
         # don't include the "operation" and "assign"
-        checks: argument,case,condition,return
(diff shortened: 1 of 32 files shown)


.golangci.yml
```diff
@@ -135,3 +135,5 @@ issues:
         - goconst
         - gocyclo
         - gomnd
+        - dupl
+        - lll
```
wass3r, line 139: suggesting to add these to be excluded from test files. would want to apply to all repos if accepted.

jbrockopp (approved):
LGTM

kneal (approved):
LGTM 🐬

PR #115, opened 2020-10-30
feat: add template to library, add map helper from compiler

davidvader:
Changes:
- Adding Template library type for use with `pipelines/*/*/templates` referenced in the pipeline endpoints feature https://github.com/go-vela/server/pull/203
- Moving map helper to yaml Template type (duplicate code used in server AND compiler) 

Why?
- YAML Template type exists here https://github.com/go-vela/types/blob/master/yaml/template.go#L15-L19 but lacks a field for HTMLURL that would greatly improve the functionality provided by`pipelines/*/*/templates`. Namely, providing the HTMLURL will allow for direct-linking to templates in the UI.
- mapFromTemplates exists in the compiler but is now being duplicated in https://github.com/go-vela/server/pull/203
https://github.com/go-vela/compiler/blob/cd7849cddf72ecd91e092ca050d7dd71d6c30f73/compiler/native/expand.go#L92-L101 this PR de-dupes and moves the code to the yaml package

### DIFF
--- a/library/template.go
+++ b/library/template.go
@@ -12,23 +12,23 @@ import (
 //
 // swagger:model Template
 type Template struct {
-	URL    *string `json:"url,omitempty"`
+	Link   *string `json:"link,omitempty"`
 	Name   *string `json:"name,omitempty"`
 	Source *string `json:"source,omitempty"`
 	Type   *string `json:"type,omitempty"`
 }
 
-// GetURL returns the URL field.
+// GetLink returns the Link field.
 //
 // When the provided Template type is nil, or the field within
 // the type is nil, it returns the zero value for the field.
-func (t *Template) GetURL() string {
-	// return zero value if Template type or URL field is nil
-	if t == nil || t.URL == nil {
+func (t *Template) GetLink() string {
+	// return zero value if Template type or Link field is nil
+	if t == nil || t.Link == nil {
 		return ""
 	}
 
-	return *t.URL
+	return *t.Link
 }
 
 // GetName returns the Name field.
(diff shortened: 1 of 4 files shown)


library/template.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"fmt"
+)
+
+// Template is the library representation of a template for a pipeline.
+//
+// swagger:model Template
+type Template struct {
+	HTMLURL *string `json:"id,omitempty"`
```
jbrockopp, line 15: I think we want to fix the `json` tag here 👍 

```suggestion
	HTMLURL *string `json:"html_url,omitempty"`
```

Also, thoughts on calling it `URL` vs `HTMLURL`?

Thinking this more closely matches what we've declared in other structs:

https://github.com/go-vela/types/blob/226d0cc72538872fc5f8f144c7d380b7ca3b4d11/library/deployment.go#L9-L22

library/template_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("String is %v, want %v", got, want)
+	}
+}
+
+// testTemplate is a test helper function to create a Template
+// type with all fields set to a fake value.
+func testTemplate() *Template {
+	u := new(Template)
```
jbrockopp, line 120: This is a bit nit-picky but could we update this to `t` for the variable name?

```suggestion
	t := new(Template)
```

jbrockopp (changes requested):

library/template.go
davidvader: went with HTMLURL to be consistent with the git payload field being `html_url`. Technically Source is a URL too? I can change it, no biggie

library/template_test.go
davidvader: good catch, updated 👍

library/template.go
davidvader: https://developer.github.com/v3/repos/contents/#get-repository-content see payloads

I like URL though. Just thought it will make more sense because the function to fetch this data will be called source.GetHTMLURL()

davidvader: updated from `HTMLURL`/`html_url` to  `URL`/`url`

jbrockopp: I hear ya 👍 

I think the real difference is that `Source` wouldn't be a fully qualified url because it's missing the protocol/scheme (`http://`, `https://`).

davidvader: agreed, good point, thank you 👍

jbrockopp (approved):
LGTM

jbrockopp (approved):
LGTM

kneal (approved):
LGTM 🐬

PR #118, opened 2020-12-02
refactor: swap yaml libraries

kneal:
This PR is towards an effort to consolidate on a single Yaml library for unmarshaling pipelines.

The plan is to work towards [goccy/go-yaml](https://github.com/goccy/go-yaml) being the core library. 

It has the desired feature we would like to use in the compiler to [enhance the error handling](https://github.com/goccy/go-yaml#51-print-customized-error-with-yaml-source-code) for users passing improper pipelines. 

**New Features:**

* All of the [validation logic](https://github.com/go-vela/compiler/blob/master/compiler/native/validate.go) that was in the compiler has not moved here. The reason is that the library gives us a new pattern for [handling errors](https://github.com/go-vela/compiler/blob/master/compiler/native/validate.go).
* The only Yaml library imported after this is merged will be https://github.com/goccy/go-yaml

fixes: https://github.com/go-vela/community/issues/9


kneal:
Screen Shot 2020-12-10 at 11 54 02 AM
This example is from a failed Go test but shows what the new validation could look like when using CLI

davidvader (approved):
LGTM

KellyMerrick (approved):
lgtm

PR #130, opened 2021-01-06
feat(yaml): allow pipeline to specify metadata.clone bool

JordanSussman:
Needed to complete https://github.com/go-vela/community/issues/96.


yaml/metadata.go
```diff
@@ -10,13 +10,21 @@ import "github.com/go-vela/types/pipeline"
 // the metadata block for a pipeline.
 // nolint:lll // jsonschema will cause long lines
 type Metadata struct {
-	Template bool `yaml:"template,omitempty" json:"template,omitempty" jsonschema:"description=Enables compiling the pipeline as a template.\nReference: https://go-vela.github.io/docs/concepts/pipeline/metadata/"`
+	Template bool  `yaml:"template,omitempty" json:"template,omitempty" jsonschema:"description=Enables compiling the pipeline as a template.\nReference: https://go-vela.github.io/docs/concepts/pipeline/metadata/"`
+	Clone    *bool `yaml:"clone,omitempty" json:"clone,omitempty" jsonschema:"description=Disables injecting the default clone step.\nReference: https://go-vela.github.io/docs/concepts/pipeline/metadata/"`
```
jbrockopp, line 14: Do you think we should update the description here?

Based off the JSON schema description for this field:

> Disables injecting the default clone step.

But below in the `metadata.ToPipeline()` function we set a default of `true` for this field.

This indicates to me that by setting the `Clone` field to `true`, it will inject the clone process.

I could see a scenario where a user thinks setting `Clone` to `true` would disable the clone process.

Thinking maybe something like:

```suggestion
	Clone    *bool `yaml:"clone,omitempty" json:"clone,omitempty" jsonschema:"default=true,description=Enables injecting the default clone process.\nReference: https://go-vela.github.io/docs/concepts/pipeline/metadata/"`
```

P.S. I switched the wording from `default clone step` to `default clone process` since it can be a stage or step.

jbrockopp (changes requested):

yaml/metadata.go
JordanSussman: Thanks! Just took the suggested changes verbatim.

jbrockopp (approved):
LGTM

wass3r (approved):

PR #133, opened 2021-01-08
fix(version): fix to upgrade semver package

wass3r:
this is needed so we can upgrade the semver package on other repos, otherwise you get this error when trying to upgrade to `semver/v3`:
```
version/version.go:42:3: cannot use v.Major() (type uint64) as type int64 in field value
version/version.go:43:3: cannot use v.Minor() (type uint64) as type int64 in field value
version/version.go:44:3: cannot use v.Patch() (type uint64) as type int64 in field value
```

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -31,11 +31,11 @@ type Version struct {
 	// Canonical represents a canonical semantic version for the application.
 	Canonical string `json:"canonical"`
 	// Major represents incompatible API changes.
-	Major int64 `json:"major"`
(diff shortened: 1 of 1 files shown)


jbrockopp (approved):
Thanks for finding that!

Although, it does make me wonder why they went to unsigned integers? 🤔

wass3r:
I guess versions can't be negative :)

KellyMerrick (approved):
lgtm!

PR #150, opened 2021-02-12
fix(schema): add 'go' as enum for template format

wass3r:
currently, the schema will mark a pipeline as invalid if you are defining `go` as the template format, eg. (taken from https://go-vela.github.io/docs/templates/):

```yaml
templates:
  - name: go
    source: github.com/octocat/hello-world/.vela/build.yml
    format: go
    type: github
```

This is, in fact, a valid option.


JordanSussman (approved):

JordanSussman:
Adding a [link](https://github.com/go-vela/compiler/blob/bab81b0cf35fb9916807b0998b5a9a44d835d259/compiler/native/expand.go#L94) to the underlying code that parses that value.

davidvader (approved):

jbrockopp (approved):
LGTM

PR #151, opened 2021-02-12
feat: add MergeEnv() for resource environment variables

jbrockopp:
Throughout the Vela repos, we have times where we need to update the environment variables for a resource.

Thus far, this has lead to us creating a helper function to accomplish this:

```go
// helper function to merge two maps together.
func appendMap(originalMap, otherMap map[string]string) map[string]string {
	for key, value := range otherMap {
		originalMap[key] = value
	}


	return originalMap
}
```

Here are a few spots where we've added this same function:

* [go-vela/compiler on `v0.7.2`](https://github.com/go-vela/compiler/blob/v0.7.2/compiler/native/environment.go#L209-L216)
* [go-vela/pkg-executor on `v0.7.2` for services](https://github.com/go-vela/pkg-executor/blob/v0.7.2/internal/service/environment.go#L70-L77)
* [go-vela/pkg-executor on `v0.7.2` for steps](https://github.com/go-vela/pkg-executor/blob/v0.7.2/internal/step/environment.go#L70-L77)

As the need to manipulate the environment variables for a resource grows, it's ideal to create a function to handle this.

The benefits of this are:

* decreased effort - no longer need to add/modify the same function in multiple repos
* increased safety - we'll now check for `nil` decreasing the potential for a `panic`
* reduced clutter of code - knowing where a small helper function is located in the repos can be a pain
* same code path - now all resources will follow the same code path for manipulating env vars throughout the repos


matt-fevold (approved):
LGTM 🐕

PR #158, opened 2021-02-25
fix: add clone to needs if not present

davidvader:
fixes the scenario where a user has multiple stages, the first stage is skipped, and the following stage(s) are executed before the `clone` stage can complete.

This bug is a race condition between the platform finishing `clone` and executing a continued stage.
  

For example, the following yaml will sometimes execute `second_stage-second_step` before clone can complete. 
Because the first stage is skipped, and the second stage provides a `needs` block, `clone` will not be added as a dependency. This code fixes that.

```yaml
version: "1"

stages:
  first_stage:
    steps:
      - name: first_step
        ruleset:
          event: pull_request
        image: golang
        commands:
          - echo "hello"
  second_stage:
    needs: [ first_stage ]
    steps:
      - name: second_step
         ruleset:
           event: push
        image: golang
        commands:
          - echo "hello"
```

related worker error:
```
error:
 unable to execute stage: 
 unable to exec step second_step: 
 Error response from daemon: 
 failed to mkdir /var/lib/docker2/volumes/DavidVader_heyvela_1394/_data/src:
 mkdir /var/lib/docker2/volumes/DavidVader_heyvela_1394/_data/src: file exists
```

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,67 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '31 2 * * 4'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
+        # Learn more:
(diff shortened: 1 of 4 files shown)


wass3r:
I think it's inferred, but just making sure. When `clone` is disabled (as implemented via https://github.com/go-vela/compiler/pull/109) it will just consider it satisfied because it's missing. And if someone implemented their own equivalent "clone" operation, they, presumably, are more likely to be more intent about adding the appropriate `needs` condition?

davidvader:
> I think it's inferred, but just making sure. When `clone` is disabled (as implemented via [go-vela/compiler#109](https://github.com/go-vela/compiler/pull/109)) it will just consider it satisfied because it's missing. And if someone implemented their own equivalent "clone" operation, they, presumably, are more likely to be more intent about adding the appropriate `needs` condition?

Yes! In fact, you can disable `clone` with metadata right now and give all stages `needs: clone` and it will have no negative impact on your pipeline because `clone` is never added as a dependent stage.
 
See these snippets from the pkg-executor:
https://github.com/go-vela/pkg-executor/blob/773d58f8732fe4fb9d7eb07fe96fc6e8d05087ba/executor/linux/build.go#L443-L444
the clone stage doesnt exist, so it doesnt get added to the stageMap
that map controls when to wait on steps
https://github.com/go-vela/pkg-executor/blob/773d58f8732fe4fb9d7eb07fe96fc6e8d05087ba/executor/linux/stage.go#L74-L78

wass3r (approved):

jbrockopp (approved):
LGTM

matt-fevold (approved):

PR #174, opened 2021-05-20
feat(database): use struct tags for custom types in gorm

jbrockopp:
Part of the effort for https://github.com/go-vela/community/issues/248

This enables us to use the following custom data types with the `gorm.io/gorm` (V2 of GORM) package:

* `github.com/go-vela/types/raw.StringSliceMap`
* `github.com/lib/pq.StringArray{}`

Currently when attempting to store a `Build{}`, `Secret{}`, `User{}` or `Worker{}` in the database, you'll see an error like:

```
[error] unsupported data type: &[]
```

This error is returned because GORM is unsure what the underlying data types should be in the database for these.

In order to resolve this, we have to explicitly tell GORM what data types to use so it can store them in the database.

The official GORM documentation on this can be found here:

https://gorm.io/docs/data_types.html

There was also an issue posted on this subject matter:

https://github.com/go-gorm/gorm/issues/1588#issuecomment-351268997

There is also a Stack Overflow post discussing this:

https://stackoverflow.com/a/64035166/14275726

### DIFF
--- a/database/build.go
+++ b/database/build.go
@@ -43,7 +43,7 @@ type Build struct {
 	Started       sql.NullInt64      `sql:"started"`
 	Finished      sql.NullInt64      `sql:"finished"`
 	Deploy        sql.NullString     `sql:"deploy"`
-	DeployPayload raw.StringSliceMap `sql:"deploy_payload"`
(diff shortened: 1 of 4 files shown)


meinenec (approved):

kneal (approved):
LGTM 🐬

PR #176, opened 2021-05-20
feat(repo): add a counter field

kneal:
PR Includes the following:

* adds a `counter` variable to the repo database struct
* adds a `counter` variable to the repo library struct

Related to: https://github.com/go-vela/community/issues/137

### DIFF
--- a/library/repo.go
+++ b/library/repo.go
@@ -20,6 +20,7 @@ type Repo struct {
 	Clone        *string `json:"clone,omitempty"`
 	Branch       *string `json:"branch,omitempty"`
 	Timeout      *int64  `json:"timeout,omitempty"`
+	Counter      *int    `json:"counter,omitempty"`
 	Visibility   *string `json:"visibility,omitempty"`
 	Private      *bool   `json:"private,omitempty"`
 	Trusted      *bool   `json:"trusted,omitempty"`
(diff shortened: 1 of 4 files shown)


library/repo.go
```diff
@@ -584,6 +611,7 @@ func (r *Repo) String() string {
   Org: %s,
   Private: %t,
   Timeout: %d,
+	Counter: %d,
```
jbrockopp, line 614: Could we fix the spacing here so it's inline with the other fields?

library/repo_test.go
```diff
@@ -290,6 +291,7 @@ func TestLibrary_Repo_String(t *testing.T) {
   Org: %s,
   Private: %t,
   Timeout: %d,
+	Counter: %d,
```
jbrockopp, line 294: Could we fix the spacing here so it's inline with the other fields?

jbrockopp (changes requested):

library/repo.go
kneal: Oh weird, my editor rendered it fine. I'll take a look

library/repo_test.go
kneal: Oh weird, my editor rendered it fine. I'll take a look

kneal: https://github.com/go-vela/types/pull/176/commits/4e63050571a1d705ee5efd1a816b6b17bd2de2f4

jbrockopp (approved):
LGTM

KellyMerrick (approved):
lgtm

PR #178, opened 2021-06-07
enhance(build): set tag env vars for deployments

jbrockopp:
This updates the `github.com/go-vela/types/library.Build.Environment()` function for the `deployment` event.

This change will ensure that if a `deployment` is created from a tag on a repo:

```sh
vela add deployment --org myOrg --repo myRepo --ref refs/tags/v0.1.0
```

That the `BUILD_TAG` and `VELA_BUILD_TAG` environment variables will be set for that build.

This is accomplished by checking if the `ref` for the build object contains the prefix `refs/tags/`.


matt-fevold (approved):

KellyMerrick (approved):
lgtm

PR #179, opened 2021-06-07
enhance(container): support slash (/) in names

jbrockopp:
This updates the `github.com/go-vela/types/pipeline.Container.Sanitize()` function.

This change will ensure that if a `service` or `step` contains a slash (`/`) in the name, Vela can support it.

This is accomplished by checking if container has a slash in the name, and if so, replacing it with a dash (`-`).

This is similar logic to how we handle other special characters such as spaces (` `), periods (`.`) or underscores (`_`).

### DIFF
--- a/pipeline/container.go
+++ b/pipeline/container.go
@@ -260,6 +260,10 @@ func (c *Container) Sanitize(driver string) *Container {
 			c.ID = strings.ReplaceAll(c.ID, " ", "-")
 		}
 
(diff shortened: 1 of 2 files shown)


matt-fevold (approved):

matt-fevold (commented):
Did we have any tests around this sanitize function?

KellyMerrick (approved):
lgtm

PR #180, opened 2021-06-08
fix: use tmp for empty name check

kneal:
PR includes the following changes:

- add a check to skips secrets that have empty names

Details:

In certain, situations the compiler can call the marshaller; sometimes when this happens with templates the secrets being passed has a secret array that contains empty secret elements. This change adds extra checking in cases a secret element exists within the secrets block that doesn't have `name:`.


jbrockopp (approved):
LGTM

JordanSussman (approved):

matt-fevold (approved):

wass3r (approved):

PR #185, opened 2021-06-18
fix(raw/map): add length check to prevent unmarshal panic

GregoryDosh:
Identified bug in https://github.com/go-vela/community/issues/300 which results in a panic due to the assumption that the strings are split into a key/value pair on `=` which isn't always true, as evidence by this contrived data example.

I'm not sure of a contrived example for the JSON case, but I'm assuming it could happen too, so I fixed it in both the YAML and JSON Unmarshal functions. If you comment out the length checks but leave the malformed step test and run it you'll see the same panic as in https://github.com/go-vela/community/issues/300


raw/map.go
```diff
@@ -54,6 +54,10 @@ func (s *StringSliceMap) UnmarshalJSON(b []byte) error {
 			// split each slice element into key/value pairs
 			kvPair := strings.SplitN(v, "=", 2)
 
+			if len(kvPair) != 2 {
```
wass3r, line 57: could add something along the lines of https://github.com/go-vela/types/blob/8396842ec2d0693ae1f97f5372debf9577426117/yaml/ulimit.go#L62 here to silence the linter. i don't think we need to make it a constant.

raw/map.go
```diff
@@ -101,6 +105,10 @@ func (s *StringSliceMap) UnmarshalYAML(unmarshal func(interface{}) error) error
...
```
wass3r, line 108: could add something along the lines of https://github.com/go-vela/types/blob/8396842ec2d0693ae1f97f5372debf9577426117/yaml/ulimit.go#L62 here to silence the linter. i don't think we need to make it a constant.

wass3r (changes requested):
couple of tweaks.. can you change the title of the PR to follow https://github.com/go-vela/types/blob/master/.github/CONTRIBUTING.md#development (it's a bit buried, but it's under the "Open a pull request!" heading).

wass3r:
Thanks for the changes. I think this makes sense. My only other suggestion would be to add a test case closer to where the fix was a applied, ie. in `/raw/map_test.go`.

GregoryDosh:
@wass3r I can take a look at updating the tests to be more localized and specific. I need to attach the debugger and figure out why panic (and the tests) weren't triggering when the code was more localized. It might shed more light on other aspects that could be spruced up.

wass3r:
@GregoryDosh sounds good - thanks! i know the current test added in the other spot covers the fix, but localizing the test will make the fixed code a bit more portable and help prevent regression. however, it _does seem_ a bit nitpicky and i'm good with approving this otherwise - thanks again for the contribution!

wass3r (approved):

GregoryDosh:
I've found a more localized test that triggers the panic in the unmarshalling for string slice maps. It feels odd though that we're hitting into that given the input data. There might be something higher level happening here that could be investigated but at least this prevents an occasional panic.

wass3r (approved):

jbrockopp (approved):
LGTM

PR #186, opened 2021-06-21
feat: add environment tag for global injection

kneal:
PR includes the following changes:

- Add `environment:` to metadata block enables controlling which containers can get global config
- Add `environment:` to the base of YAML allows setting what config is globally injected
- New helper function on metadata struct for compiler injection control

Story for additionally context https://github.com/go-vela/community/issues/304

Additional related PRs:
-

### DIFF
--- a/yaml/metadata.go
+++ b/yaml/metadata.go
@@ -4,11 +4,7 @@
 
 package yaml
 
-import (
-	"fmt"
-
-	"github.com/go-vela/types/pipeline"
-)
+import "github.com/go-vela/types/pipeline"
 
 type (
 	// Metadata is the yaml representation of
@@ -73,10 +69,9 @@ func (m *Metadata) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	}
 
 	// overwrite existing metadata environment details
-	m = (*Metadata)(metadata)
-
-	fmt.Println("m ENV: ", m.Environment)
-	fmt.Println("m CLONE", *m.Clone)
+	m.Template = metadata.Template
+	m.Clone = metadata.Clone
+	m.Environment = metadata.Environment
 
 	return nil
 }
--- a/yaml/metadata.go
+++ b/yaml/metadata.go
@@ -4,15 +4,27 @@
 
 package yaml
 
-import "github.com/go-vela/types/pipeline"
+import (
+	"fmt"
 
-// Metadata is the yaml representation of
-// the metadata block for a pipeline.
-// nolint:lll // jsonschema will cause long lines
-type Metadata struct {
-	Template bool  `yaml:"template,omitempty" json:"template,omitempty" jsonschema:"description=Enables compiling the pipeline as a template.\nReference: https://go-vela.github.io/docs/reference/yaml/metadata/#the-template-tag"`
-	Clone    *bool `yaml:"clone,omitempty" json:"clone,omitempty" jsonschema:"default=true,description=Enables injecting the default clone process.\nReference: https://go-vela.github.io/docs/reference/yaml/metadata/#the-clone-tag"`
-}
+	"github.com/go-vela/types/pipeline"
+)
+
+type (
+	// Metadata is the yaml representation of
+	// the metadata block for a pipeline.
+	// nolint:lll // jsonschema will cause long lines
+	Metadata struct {
+		Template    bool     `yaml:"template,omitempty" json:"template,omitempty" jsonschema:"description=Enables compiling the pipeline as a template.\nReference: https://go-vela.github.io/docs/reference/yaml/metadata/#the-template-tag"`
+		Clone       *bool    `yaml:"clone,omitempty" json:"clone,omitempty" jsonschema:"default=true,description=Enables injecting the default clone process.\nReference: https://go-vela.github.io/docs/reference/yaml/metadata/#the-clone-tag"`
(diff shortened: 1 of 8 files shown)


yaml/metadata.go
```diff
@@ -4,15 +4,25 @@
...
+	Metadata struct {
+		Template    bool     `yaml:"template,omitempty" json:"template,omitempty" jsonschema:"description=Enables compiling the pipeline as a template.\nReference: https://go-vela.github.io/docs/reference/yaml/metadata/#the-template-tag"`
+		Clone       *bool    `yaml:"clone,omitempty" json:"clone,omitempty" jsonschema:"default=true,description=Enables injecting the default clone process.\nReference: https://go-vela.github.io/docs/reference/yaml/metadata/#the-clone-tag"`
+		Environment []string `yaml:"environment,omitempty" json:"environment,omitempty" jsonschema:"default=true,description=Controls which containers processes can have global env injected.\nReference: https://go-vela.github.io/docs/reference/yaml/metadata/#the-environment-tag"`
+	}
+
+	// helper type that allows the unmarshaler interface
+	// to add default values back into metadata. Using the
+	// Metadata type directly will result in a reflection error.
+	_metadata Metadata
```
jbrockopp, line 24: What are your thoughts on declaring an anonymous struct instead of this?

In case it's not clear what I'm proposing, I'll drop a comment later on where this type is used 👍

yaml/metadata.go
```diff
@@ -23,8 +33,45 @@ func (m *Metadata) ToPipeline() *pipeline.Metadata {
...
+		}
+	}
+
+	return false
+}
+
+// UnmarshalYAML implements the Unmarshaler interface for the Metadata type.
+func (m *Metadata) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	// metadata we try unmarshalling to
+	metadata := new(_metadata)
```
jbrockopp, line 59: Related to my earlier comment - we could declare an anonymous struct here instead.

This would follow the conventions and patterns already established for other types.

Specifically where we override the YAML unmarshaling behavior:

https://github.com/go-vela/types/blob/d9de354942deb1a8104aa1ba451b88e342555199/yaml/ruleset.go#L56-L63

https://github.com/go-vela/types/blob/d9de354942deb1a8104aa1ba451b88e342555199/yaml/ruleset.go#L125-L134

jbrockopp (changes requested):

yaml/metadata.go
kneal: Yeah, I can swap around to that! I honestly wasn't sure how to handle it and didn't recognize we had that pattern. Easy change.

yaml/metadata.go
```diff
@@ -4,15 +4,25 @@
...
+)
+
+type (
+	// Metadata is the yaml representation of
+	// the metadata block for a pipeline.
+	// nolint:lll // jsonschema will cause long lines
+	Metadata struct {
+		Template    bool     `yaml:"template,omitempty" json:"template,omitempty" jsonschema:"description=Enables compiling the pipeline as a template.\nReference: https://go-vela.github.io/docs/reference/yaml/metadata/#the-template-tag"`
+		Clone       *bool    `yaml:"clone,omitempty" json:"clone,omitempty" jsonschema:"default=true,description=Enables injecting the default clone process.\nReference: https://go-vela.github.io/docs/reference/yaml/metadata/#the-clone-tag"`
+		Environment []string `yaml:"environment,omitempty" json:"environment,omitempty" jsonschema:"default=true,description=Controls which containers processes can have global env injected.\nReference: https://go-vela.github.io/docs/reference/yaml/metadata/#the-environment-tag"`
```
wass3r, line 18: this is really an enum, right? can we add that to the `jsonschema` tag? example from elsewhere: https://github.com/go-vela/types/blob/8396842ec2d0693ae1f97f5372debf9577426117/yaml/ruleset.go#L20

we also probably don't want `default=true` - might be copy pasta error

wass3r (changes requested):

yaml/metadata.go
wass3r: sorry, not sure what i was thinking - it's more similar to https://github.com/go-vela/types/blob/84b1e1d2c2bcde72e867d5f0dd945b660db37453/yaml/ruleset.go#L30 and we custom handle that in the schema generator.

at the very least, `default=true` should be removed in the tag.

yaml/metadata.go
kneal: Belive this is what you were looking for:
https://github.com/go-vela/types/pull/186/commits/cac25f3ed967c1b8286db45541162783b347f580

yaml/metadata.go
kneal: What does the `default=true` mean? Because technically if a user doesn't provide the field we inject

`environment: [ steps, services, secrets ]` into the YAML block. Simplified the compiler code to do that because then we can assume the field is never empty or nil

jbrockopp (approved):
LGTM

yaml/metadata.go
wass3r: it says the default value is `true`

kneal: Rmoved https://github.com/go-vela/types/pull/186/commits/76f522399120283e163186af79a40d67d662b1eb

wass3r (approved):

jbrockopp (approved):
LGTM

PR #198, opened 2021-09-04
fix: reference correct secret parts in errors

wass3r:
fixes https://github.com/go-vela/community/issues/387

When we call `.ParseOrg`, we pass in the current repo's information (eg. https://github.com/go-vela/pkg-executor/blob/57b690402cf77554746eebe9e870ae7a3ec1e37d/executor/linux/secret.go#L196 - passing in current repo's org).

When a user provides a secret from a different org via the `key` property, the error prints the current org the user is trying to use the secret from instead of the org that they are referencing in the `key`.

for example:

Assuming the following pipeline is used in `github.com/go-vela/test`:
```yaml
version: "1"

steps:
  - name: test
    image: alpine:latest
    pull: true
    command:
      - echo hi

secrets:
  - name: org-secret
    key: wass3r/secret
    type: org
    engine: native
```

The error would print:
```
error:invalid organization in key: go-vela
```
instead of:
```
error:invalid organization in key: wass3r
```

### DIFF
--- a/pipeline/secret.go
+++ b/pipeline/secret.go
@@ -124,7 +124,7 @@ func (s *Secret) ParseOrg(org string) (string, string, error) {
 
 	// check if the org provided matches what we expect
 	if !strings.EqualFold(parts[0], org) {
-		return "", "", fmt.Errorf("%s: %s ", ErrInvalidOrg, org)
+		return "", "", fmt.Errorf("%s: %s ", ErrInvalidOrg, parts[0])
 	}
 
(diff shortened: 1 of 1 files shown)


wass3r:
one thought.. should we print the whole supplied `key` value instead?

matt-fevold (approved):
LGTM

matt-fevold:
> one thought.. should we print the whole supplied key value instead?

Not sure 🤔 I think what you've got already should help reduce confusion a lot.

PR #199, opened 2021-09-24
enhance(container): enable setting user

JordanSussman:
Need for https://github.com/go-vela/community/issues/344.

### DIFF
--- a/pipeline/container.go
+++ b/pipeline/container.go
@@ -50,6 +50,7 @@ type (
 		Secrets     StepSecretSlice   `json:"secrets,omitempty"     yaml:"secrets,omitempty"`
 		Ulimits     UlimitSlice       `json:"ulimits,omitempty"     yaml:"ulimits,omitempty"`
 		Volumes     VolumeSlice       `json:"volumes,omitempty"     yaml:"volumes,omitempty"`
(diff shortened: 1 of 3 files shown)


jbrockopp (changes requested):
The code you have written LGTM! 🎉 

However, I noticed that the story linked refers to enabling this for both `steps` and `services`.

If I'm following correctly, I believe this adds YAML support for the `user` key for a step, but not for a service.

Should there also be code here for adding a `User` string field to the `Service` struct?

https://github.com/go-vela/types/blob/c8edeac9eaa00634f91e3a58bc754f4f5b45230d/yaml/service.go#L21-L32

JordanSussman:
> The code you have written LGTM! 🎉
> 
> However, I noticed that the story linked refers to enabling this for both `steps` and `services`.
> 
> If I'm following correctly, I believe this adds YAML support for the `user` key for a step, but not for a service.
> 
> Should there also be code here for adding a `User` string field to the `Service` struct?
> 
> https://github.com/go-vela/types/blob/c8edeac9eaa00634f91e3a58bc754f4f5b45230d/yaml/service.go#L21-L32

Good catch! The `user` parameter was inadvertently left out of the original commit.

jbrockopp (approved):
LGTM

wass3r (approved):

PR #200, opened 2021-10-11
kubernetes: sanitize container names (make lowercase and dns-safe)

cognifloyd:
Kubernetes requires lowercase, dns-safe (<=63 chars long) container names.
So, this modifies Sanitize() to ensure we meet those requirements.

Without this, I run into errors if I use any uppercase characters in step names, or if the build + stage + step names are longer than 63 characters (which is really easy to do).


cognifloyd:
I just pushed an edit to silence golangci-lint.

PR #206, opened 2021-10-29
feat(secrets): add secret metadata

ecrupper:
Adding more metadata for secrets including the created time, user that created the secret, last updated time, and the build id of the last build where it was used

### DIFF
--- a/database/secret.go
+++ b/database/secret.go
@@ -54,6 +54,11 @@ type Secret struct {
 	Images       pq.StringArray `sql:"images" gorm:"type:varchar(1000)"`
 	Events       pq.StringArray `sql:"events" gorm:"type:varchar(1000)"`
 	AllowCommand sql.NullBool   `sql:"allow_command"`
+	CreatedAt    sql.NullString `sql:"created_at"`
+	CreatedBy    sql.NullInt64  `sql:"created_by"`
+	UpdatedAt    sql.NullString `sql:"updated_at"`
+	UpdatedBy    sql.NullInt64  `sql:"updated_by"`
+	LastBuildID  sql.NullInt64  `sql:"last_build_id"`
 }
 
 // Decrypt will manipulate the existing secret value by
@@ -149,6 +154,31 @@ func (s *Secret) Nullify() *Secret {
 		s.Type.Valid = false
 	}
 
+	// check if the CreatedAt field should be false
+	if len(s.CreatedAt.String) == 0 {
+		s.CreatedAt.Valid = false
+	}
+
+	// check if the CreatedBy field should be false
+	if s.CreatedBy.Int64 == 0 {
+		s.CreatedBy.Valid = false
+	}
+
+	// check if the UpdatedAt field should be false
+	if len(s.UpdatedAt.String) == 0 {
+		s.UpdatedAt.Valid = false
+	}
+
+	// check if the UpdatedBy field should be false
+	if s.UpdatedBy.Int64 == 0 {
+		s.UpdatedBy.Valid = false
+	}
+
+	// check if the LastBuildID field should be false
+	if s.LastBuildID.Int64 == 0 {
+		s.LastBuildID.Valid = false
+	}
+
 	return s
 }
 
@@ -167,6 +197,11 @@ func (s *Secret) ToLibrary() *library.Secret {
 	secret.SetImages(s.Images)
 	secret.SetEvents(s.Events)
 	secret.SetAllowCommand(s.AllowCommand.Bool)
+	secret.SetCreatedAt(s.CreatedAt.String)
+	secret.SetCreatedBy(s.CreatedBy.Int64)
+	secret.SetUpdatedAt(s.UpdatedAt.String)
+	secret.SetUpdatedBy(s.UpdatedBy.Int64)
+	secret.SetLastBuildID(s.LastBuildID.Int64)
 
 	return secret
 }
@@ -249,6 +284,11 @@ func SecretFromLibrary(s *library.Secret) *Secret {
 		Images:       pq.StringArray(s.GetImages()),
 		Events:       pq.StringArray(s.GetEvents()),
 		AllowCommand: sql.NullBool{Bool: s.GetAllowCommand(), Valid: true},
+		CreatedAt:    sql.NullString{String: s.GetCreatedAt(), Valid: true},
+		CreatedBy:    sql.NullInt64{Int64: s.GetCreatedBy(), Valid: true},
+		UpdatedAt:    sql.NullString{String: s.GetUpdatedAt(), Valid: true},
+		UpdatedBy:    sql.NullInt64{Int64: s.GetUpdatedBy(), Valid: true},
+		LastBuildID:  sql.NullInt64{Int64: s.GetLastBuildID(), Valid: true},
 	}
 
 	return secret.Nullify()
--- a/database/secret.go
+++ b/database/secret.go
@@ -58,7 +58,6 @@ type Secret struct {
 	CreatedBy    sql.NullInt64  `sql:"created_by"`
 	UpdatedAt    sql.NullInt64  `sql:"updated_at"`
 	UpdatedBy    sql.NullInt64  `sql:"updated_by"`
-	LastBuildID  sql.NullInt64  `sql:"last_build_id"`
 }
 
 // Decrypt will manipulate the existing secret value by
@@ -174,11 +173,6 @@ func (s *Secret) Nullify() *Secret {
 		s.UpdatedBy.Valid = false
 	}
 
-	// check if the LastBuildID field should be false
-	if s.LastBuildID.Int64 == 0 {
-		s.LastBuildID.Valid = false
-	}
-
 	return s
 }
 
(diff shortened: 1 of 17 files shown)


library/secret.go
```diff
@@ -26,6 +26,10 @@ type Secret struct {
 	Images       *[]string `json:"images,omitempty"`
 	Events       *[]string `json:"events,omitempty"`
 	AllowCommand *bool     `json:"allow_command,omitempty"`
+	CreateTime   *string   `json:"create_time,omitempty"`
+	CreatedBy    *string   `json:"created_by,omitempty"`
```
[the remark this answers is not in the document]
ecrupper: Sounds great! Just implemented that with last commit

library/secret.go
```diff
@@ -26,6 +26,10 @@ type Secret struct {
 	Images       *[]string `json:"images,omitempty"`
 	Events       *[]string `json:"events,omitempty"`
 	AllowCommand *bool     `json:"allow_command,omitempty"`
+	CreateTime   *string   `json:"create_time,omitempty"`
```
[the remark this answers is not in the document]
ecrupper: Yes that looks much more consistent - fixed in latest commit

library/secret.go
```diff
@@ -26,6 +26,10 @@ type Secret struct {
...
+	CreatedBy    *string   `json:"created_by,omitempty"`
+	UpdateTime   *string   `json:"update_time,omitempty"`
```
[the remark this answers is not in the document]
ecrupper: Done!

library/secret.go
```diff
@@ -26,6 +26,10 @@ type Secret struct {
...
+	LastBuildID  *int64    `json:"last_build,omitempty"`
 }
```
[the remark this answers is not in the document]
ecrupper: Implemented!

wass3r (changes requested):
Thanks for the PR! Few additional things:

1. Assuming that `CreatedBy` and `UpdatedBy` will reference User IDs, we should match the type, ie. change from `string` to `int64`.
1. I think we'll need to add these new fields to `/database/secret.go` as well
1. No action for this PR, but just a reminder that we will need to have an accompanying PR that adds the fields in https://github.com/go-vela/server/tree/master/database as well

database/secret_test.go
```diff
@@ -8,10 +8,15 @@ import (
...
 	"reflect"
 	"testing"
+	"time"
 
 	"github.com/go-vela/types/library"
 )
 
+var currentTime time.Time = time.Now()
+var tsCreate string = currentTime.Format(time.UnixDate)
+var tsUpdate string = currentTime.Add(time.Hour * 1).Format(time.UnixDate)
```
jbrockopp, line 18: I think you can skip providing the underlying type of the variable if you're assigning it a value.

Also, this one is a bit nit-picky but could we update this to be a single `var` block?

```suggestion
var (
    currentTime = time.Now()
    tsCreate = currentTime.Format(time.UnixDate)
    tsUpdate = currentTime.Add(time.Hour * 1).Format(time.UnixDate)
)
```

jbrockopp (changes requested):
Thanks for updating this with the previous changes! 🎉 

One more thing I noticed as I was reviewing:

database/secret_test.go
ecrupper: Yeah for sure! On this, I was informed that we are recording timestamps in our build object as ints. [https://github.com/go-vela/types/blob/911ffbea2e427883ccca2cc3f3769fdae75e0946/database/build.go#L42-L43](url). Example of utilization: [https://github.com/go-vela/server/blob/88520ca76a0856316b6c17d15d8305198d603735/api/build.go#L141](url). I will update that to be consistent.

kneal:
Just a clarifying question, `LastBuildID` is that going to be the last build to pull the secret?

ecrupper:
> Just a clarifying question, `LastBuildID` is that going to be the last build to pull the secret?

Yes that is the current plan

ecrupper:
> Cool! Just checking, I generally recommend linking the issue in the body of the PR. Helps with that extra context and create circular deps to the PR to issue etc.

Absolutely! It was an internal request that wasn't duplicated to the community issue board yet. Here it is for reference: [https://github.com/go-vela/community/issues/428](url)

library/secret.go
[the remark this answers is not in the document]
ecrupper: Hilariously, `update_at` seems to be a Postgres trigger keyword (at least in the testing DB), which causes it to override any user input and instead put the .Now() timestamp in the field. Any thoughts on a different name? `CreatedTs` and `UpdatedTs` maybe? Open to suggestions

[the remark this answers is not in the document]
jbrockopp: I think this is actually related to the Golang library we're using for this:

https://github.com/go-gorm/gorm

I found these docs which appear to describe the issue you're referencing:

https://gorm.io/docs/models.html#Creating-Updating-Time-Unix-Milli-Nano-Seconds-Tracking

This issue might be related?

https://github.com/go-gorm/gorm/issues/4827

ecrupper:
[https://github.com/go-vela/server/pull/526](url)

Draft PR in the go-vela/server that would accompany this one

database/secret.go
```diff
@@ -54,6 +54,10 @@ type Secret struct {
 	Images       pq.StringArray `sql:"images" gorm:"type:varchar(1000)"`
 	Events       pq.StringArray `sql:"events" gorm:"type:varchar(1000)"`
 	AllowCommand sql.NullBool   `sql:"allow_command"`
+	CreatedAt    sql.NullInt64  `sql:"created_at"`
+	CreatedBy    sql.NullInt64  `sql:"created_by"`
+	UpdatedAt    sql.NullInt64  `sql:"updated_at"`
+	UpdatedBy    sql.NullInt64  `sql:"updated_by"`
```
jbrockopp, line 60: Is it safe to assume that `CreatedBy` and `UpdatedBy` are `int64` types since they'll store the user's ID?

https://github.com/go-vela/types/blob/911ffbea2e427883ccca2cc3f3769fdae75e0946/database/user.go#L48-L58

IIRC, they were `string` types earlier which is why I'm asking.

library/secret.go
```diff
@@ -227,6 +235,58 @@ func (s *Secret) GetAllowCommand() bool {
...
+func (s *Secret) GetCreatedBy() int64 {
+	// return zero value if Secret type or CreatedBy field is nil
+	if s == nil || s.CreatedBy == nil {
+		return 0
+	}
+
+	return *s.CreatedBy
+}
+
+// GetUpdateAt returns the UpdateAt field.
```
jbrockopp, line 264: typo

```suggestion
// GetUpdatedAt returns the UpdatedAt field.
```

library/secret.go
```diff
@@ -227,6 +235,58 @@ func (s *Secret) GetAllowCommand() bool {
...
+//
+// When the provided Secret type is nil, or the field within
+// the type is nil, it returns the zero value for the field.
+func (s *Secret) GetUpdatedAt() int64 {
+	// return zero value if Secret type or UpdateAt field is nil
```
jbrockopp, line 269: typo

```suggestion
	// return zero value if Secret type or UpdatedAt field is nil
```

jbrockopp (changes requested):

database/secret.go
ecrupper: Yes after taking a look at that User struct, I thought it best to change those to int64 types. Originally I thought it might be stored as a name, but I believe int64 is best.

library/secret.go
ecrupper: Fixed! Thanks

library/secret.go
ecrupper: Fixed! Thanks

database/secret.go
jbrockopp: @ecrupper ok thanks for the extra information 👍 

Out of curiosity, could you elaborate on why `int64` is the approach we're choosing?

I think I tend to agree that from the data storage perspective, `int64` is better than `string` for the database.

However, my concern with this approach has to do with data access and usage patterns.

Today, there isn't a great way to query a user by their `ID` in Vela.

So if I'm looking at a secret with the UI/CLI/API, I would only be able to see the user's `ID` who created/updated it.

To actually see what user created/updated the secret, I'd have to ask an admin "Who is user <id> in Vela?"

Perhaps one way to address these concerns is to add a new endpoint to Vela which allows looking a user up by their ID?

Otherwise we could consider storing the user's `Name` rather than `ID` in the `secret` object?

What are your thoughts on this?

ecrupper: That's a great point. Originally I was thinking in terms of consistency (if there's a User table with IDs as their primary key, it would make sense to have any reference to a User  in another table be their ID). However, after considering the end user as you said, I agree that it should be a string. Not only would it save the UI an extra API call, but considering the CLI, running another command to find the name associated with a User ID seems tedious. I'll make that change!

jbrockopp (approved):
LGTM

database/secret.go
wass3r: fwiw, i proposed user id initially to be consistent with the repo object in the `repos` table where we use `int64` for user id. another instance where user _name_ could be beneficial when trying to figure out who "owns" a repo.

in the future we might consider embedding the user struct in the response similar to how github does it for some of it's objects (eg. check response here: https://docs.github.com/en/rest/reference/issues#get-an-issue).

(waiting for @JordanSussman to chime in with the obligatory GraphQL commentary 😄)

thanks for the work and revisions @ecrupper !

wass3r (approved):

database/secret.go
JordanSussman: > (waiting for @JordanSussman to chime in with the obligatory GraphQL commentary 😄)

Don't tempt me a good time! Would love to add a v2 API with graphql and/or parameters to allow parsed output.

graphql

kneal (approved):
LGTM 🐬

PR #207, opened 2021-11-05
feat(environment): add pusher name and email for tag events

ecrupper:
Currently, when a user drafts a release, the build author remains the person who last committed to the repository. We want to keep this, but we also want to know who tagged the release. Therefore, these additional fields in tandem with the server setting these values from GitHub will expose that information as environment variables.

### DIFF
--- a/library/build.go
+++ b/library/build.go
@@ -45,6 +45,8 @@ type Build struct {
 	Host          *string             `json:"host,omitempty"`
 	Runtime       *string             `json:"runtime,omitempty"`
 	Distribution  *string             `json:"distribution,omitempty"`
+	PusherName    *string             `json:"pusher_name,omitempty"`
+	PusherEmail   *string             `json:"pusher_email,omitempty"`
 }
 
 // Environment returns a list of environment variables
@@ -117,7 +119,6 @@ func (b *Build) Environment(workspace, channel string) map[string]string {
 	if strings.EqualFold(b.GetEvent(), constants.EventDeploy) {
 		// capture the deployment target
 		target := ToString(b.GetDeploy())
-
 		// add the deployment target to the list
 		envs["VELA_BUILD_TARGET"] = target
 		envs["VELA_DEPLOYMENT"] = target
@@ -131,6 +132,8 @@ func (b *Build) Environment(workspace, channel string) map[string]string {
 			// add the tag reference to the list
 			envs["BUILD_TAG"] = tag
 			envs["VELA_BUILD_TAG"] = tag
+			envs["VELA_BUILD_TAG_AUTHOR"] = b.GetPusherName()
+			envs["VELA_BUILD_TAG_AUTHOR_EMAIL"] = b.GetPusherEmail()
 		}
 
 		// add payload data to the list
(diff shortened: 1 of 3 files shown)


library/build.go
```diff
@@ -542,6 +547,22 @@ func (b *Build) GetDistribution() string {
 	return *b.Distribution
 }
 
+func (b *Build) GetPusherName() string {
+	if b == nil || b.PusherName == nil {
+		return ""
+	}
+
+	return *b.PusherName
+}
```
jbrockopp, line 556: Looks like comments are missing here.

To show an example of what those comments may look like:

https://github.com/go-vela/types/blob/911ffbea2e427883ccca2cc3f3769fdae75e0946/library/build.go#L415-L426

library/build.go
```diff
@@ -542,6 +547,22 @@ func (b *Build) GetDistribution() string {
...
+
+func (b *Build) GetPusherEmail() string {
+	if b == nil || b.PusherEmail == nil {
+		return ""
+	}
+
+	return *b.PusherEmail
+}
```
jbrockopp, line 564: Looks like comments are missing here.

To show an example of what those comments may look like:

https://github.com/go-vela/types/blob/911ffbea2e427883ccca2cc3f3769fdae75e0946/library/build.go#L428-L439

library/build.go
```diff
@@ -919,6 +940,22 @@ func (b *Build) SetDistribution(v string) {
 	b.Distribution = &v
 }
 
+func (b *Build) SetPusherName(v string) {
+	if b == nil {
+		return
+	}
+
+	b.PusherName = &v
+}
```
jbrockopp, line 949: Looks like comments are missing here.

To show an example of what those comments may look like:

https://github.com/go-vela/types/blob/911ffbea2e427883ccca2cc3f3769fdae75e0946/library/build.go#L792-L803

library/build.go
```diff
@@ -919,6 +940,22 @@ func (b *Build) SetDistribution(v string) {
...
+
+func (b *Build) SetPusherEmail(v string) {
+	if b == nil {
+		return
+	}
+
+	b.PusherEmail = &v
+}
```
jbrockopp, line 957: Looks like comments are missing here.

To show an example of what those comments may look like:

https://github.com/go-vela/types/blob/911ffbea2e427883ccca2cc3f3769fdae75e0946/library/build.go#L805-L816

jbrockopp (changes requested):
Requested some changes but I also have a question.

We're adding two new fields to the `library.Build{}` type:

* `PusherName`
* `PusherEmail`

Do we also need to add these fields to the `database.Build{}` type?

https://github.com/go-vela/types/blob/911ffbea2e427883ccca2cc3f3769fdae75e0946/database/build.go#L32-L63

ecrupper:
> Requested some changes but I also have a question.
> 
> We're adding two new fields to the `library.Build{}` type:
> 
> * `PusherName`
> * `PusherEmail`
> 
> Do we also need to add these fields to the `database.Build{}` type?
> 
> https://github.com/go-vela/types/blob/911ffbea2e427883ccca2cc3f3769fdae75e0946/database/build.go#L32-L63

Yeah that was something I was discussing with @davidvader . Since we are adding the field to the library type, it would make sense to be consistent with the database. However, since their use is purely for creating some environment variables, I figured I'd get some more feedback on whether it would be wise to include it in the database. Looking at code such as this: [https://github.com/go-vela/types/blob/911ffbea2e427883ccca2cc3f3769fdae75e0946/library/build.go#L132-L133](url) where we aren't ever storing those values on our end, I thought it would be worth discussing. Further, taking a look at @JordanSussman 's comments on the related server PR for this (draft - [https://github.com/go-vela/server/pull/527](url)), it may not even be the right approach at all. Perhaps using sender, which we already store, is the right way to go about this. However, finding the email may be tricky. Would love to hear more thoughts!

ecrupper:
Closing this PR in favor of utilizing the `sender` field and not including the `TAGGER_EMAIL` since we cannot reliable get that information from GitHub webhook.

PR #208, opened 2021-11-08
fix: annotate payload for swagger

wass3r:
part of https://github.com/go-vela/community/issues/432

`Version` is a payload expected for the `<api-url>/version` endpoint.


jbrockopp (approved):
LGTM

davidvader (approved):

PR #210, opened 2021-11-11
fix(secret)!: remove implicit repo secrets and clean up secrets testing

GregoryDosh:
To fix https://github.com/go-vela/community/issues/158 this removes the implicit repo secrets along with adding a bit more robust error checking and error wrapping.

There were a few checks for things such as string paths being exactly 2 parts long, but the splits before that were already enforcing the arbitrary break for things that are too long, so I removed a few checks that were doing nothing. I also added in a few checks for other situations and made the error checking a little bit more robust.


pipeline/secret.go
```diff
@@ -105,26 +105,21 @@ func (s *Secret) ParseOrg(org string) (string, string, error) {
...
-
-	// check if a path was provided
-	if !strings.Contains(path, "/") {
-		return "", "", fmt.Errorf("%s: %s ", ErrInvalidPath, path)
+		return "", "", fmt.Errorf("%w: %s", ErrInvalidEngine, s.Engine)
 	}
 
 	// split the full path into parts
-	parts := strings.SplitN(path, "/", 2)
+	parts := strings.Split(path, "/")
```
wass3r, line 112: I'm not sure this will work, because you can have secrets with a name like `docker/username` so an org-level secret's key could be `<org>/docker/username`. At least that's a possibility today.

if you add a test case for that you will see that this change will break those type of secrets.

wass3r (commented):
just to clarify, this would be a breaking change for anyone that has secrets defined with the implicit method?

also left a comment for consideration

pipeline/secret.go
GregoryDosh: Ahah, I hadn't realized there was a possibility to have a sort of "nested" secret at the org level. Need more tests in place to demonstrate that behavior :-P

GregoryDosh: @wass3r Is it true to say that secret names can contain any arbitrary length of forward slashes in it? Is there a restricted set of characters somewhere for the naming?

wass3r: yea, no limitation that i'm aware of.

cognifloyd:
I forget which characters I used, but there are characters that don't work. You can name the secrets whatever, but retrieving them doesn't work when they have special chars

GregoryDosh:
> I forget which characters I used, but there are characters that don't work. You can name the secrets whatever, but retrieving them doesn't work when they have special chars

If you recall which ones didn't work, could you let me know? I was just trying with some emoji and it seemed okayish. I think the limit might be in the shell support of the names in environmental variables from what I've read: https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html
Seemingly anything is valid so long as it isn't `=` or `NUL`.  I'd love to find edge cases that cause problems and get testing in place for 'em

wass3r:
Looks like some other merged change created conflicts here :(

pipeline/secret.go
```diff
@@ -119,12 +122,22 @@ func (s *Secret) ParseOrg(org string) (string, string, error) {
...
+		return "", "", fmt.Errorf("%w: %s ", ErrInvalidOrg, parts[0])
+	}
+
+	// check if path segments empty
+	if len(parts[1]) == 0 {
+		return "", "", fmt.Errorf("%w: %s ", ErrInvalidPath, path)
+	}
+
+	// secret names can't contain restricted characters.
+	if strings.ContainsAny(s.Name, constants.SecretRestrictedCharacters) {
```
wass3r, line 139: do we really want this? wouldn't this mean that you can't have a name with an `x` or a `1`? do we need to create an array for disallowed character(s) sequences instead?

GregoryDosh: I'll need to double check. I think it was supposed to actually be `\u0000` which is the nul character but I think when I was copying and pasting I pasted the wrong value, good catch. But it should only be the 2 characters the "null" control character and =. Technically I don't think you're able to put null characters in yaml but if in the future it changes or I haven't checked Starlark either.

GregoryDosh: And `ContainsAny` will iterate over the string and if any character in the secondary string is part of the first it'll error. It doesn't have to match exactly and if you name a secret with an = somewhere in the middle (I don't know why you would) it would caused issues with assignment and the values containing part of the name in the value.

```yaml
- name: some=password
```
->
```sh
export some=password=value
```
So the env var name would be injected eventually as `$SOME`

jbrockopp (dismissed):
LGTM

KellyMerrick:
Looks like there are some merge conflicts.

KellyMerrick (approved):
lgtm

jbrockopp (approved):
LGTM

PR #211, opened 2021-11-12
fix: update magic number lint issue

ecrupper:
By creating some constants related to number parsing and formatting, I was able to knock out all of the linting errors caught by Review Dog. Totally open to other suggestions on fixing the issues - just figured it would be nice to start somewhere.

### DIFF
--- a/constants/event.go
+++ b/constants/event.go
@@ -20,4 +20,7 @@ const (
 
 	// EventComment defines the event type for comments added to a pull request.
 	EventComment = "comment"
+
+	// EventRepository defines the event type for repository actions such as delete and rename.
+	EventRepository = "repository"
 )
--- a/constants/event.go
+++ b/constants/event.go
@@ -20,7 +20,4 @@ const (
 
 	// EventComment defines the event type for comments added to a pull request.
 	EventComment = "comment"
-
-	// EventRepository defines the event type for repository actions such as delete and rename.
-	EventRepository = "repository"
 )
--- a/constants/number.go
+++ b/constants/number.go
@@ -1,0 +1,23 @@
+// Copyright (c) 2021 Target Brands, Inc. All rights reserved.
+//
+// Use of this source code is governed by the LICENSE file in this repository.
+
+package constants
+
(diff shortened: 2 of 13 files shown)


kneal:
Overall, the feedback from me would be, I'm not a big fan of super generic constants. It tends to go against [domain-driven design](https://www.calhoun.io/moving-towards-domain-driven-design-in-go/) where you try and keep relative bits of information grouped together. That type of design philosophy rarely works out perfectly so you always have to make compromises but overall I think it's a very good rule of them when working on Go packages.   

Creating super generic constants that can be imported anywhere literally introduces them imported anywhere which means now you're supporting compatibility requirements anywhere those constants are imported. Again, sometimes this isn't a big deal but creating more touchpoints in my view isn't always the most ideal solution for solving this particular linting issue. 

My vote would be if we want to get rid of the linter errors we create local constants to each package or event think about just removing that linter type altogether. I'm personally not aware of how much value it's actually providing us at this point.

ecrupper:
> Overall, the feedback from me would be, I'm not a big fan of super generic constants. It tends to go against [domain-driven design](https://www.calhoun.io/moving-towards-domain-driven-design-in-go/) where you try and keep relative bits of information grouped together. That type of design philosophy rarely works out perfectly so you always have to make compromises but overall I think it's a very good rule of them when working on Go packages.
> 
> Creating super generic constants that can be imported anywhere literally introduces them imported anywhere which means now you're supporting compatibility requirements anywhere those constants are imported. Again, sometimes this isn't a big deal but creating more touchpoints in my view isn't always the most ideal solution for solving this particular linting issue.
> 
> My vote would be if we want to get rid of the linter errors we create local constants to each package or event think about just removing that linter type altogether. I'm personally not aware of how much value it's actually providing us at this point.

Thanks for the input! I will work on decentralizing the constants. Since the linter is only tripping on these couple functions, I think addressing them won't be too much of a headache.

ecrupper:
Closing this PR to work on alternate solutions for linting errors

PR #212, opened 2021-11-15
feat: add tag author environment variable

ecrupper:
Currently, whenever someone tags a release for a project, the `VELA_BUILD_AUTHOR` field is the last person to commit to the project. Since we probably want to keep that information, I am creating a new environment variable for the tag author. Here is the related issue: [https://github.com/go-vela/community/issues/436](url). Below is a screen shot of the issue on Vela UI
Screen Shot 2021-11-15 at 10 09 43 AM

Unfortunately, since the `pusher` field from GitHub is unreliable, and the `sender` field only includes the name, the tagger email will not be a part of this feature.


library/build.go
```diff
@@ -131,6 +131,7 @@ func (b *Build) Environment(workspace, channel string) map[string]string {
 			// add the tag reference to the list
 			envs["BUILD_TAG"] = tag
 			envs["VELA_BUILD_TAG"] = tag
+			envs["VELA_BUILD_TAG_AUTHOR"] = b.GetSender()
```
jbrockopp, line 134: AFAICT, this `VELA_BUILD_TAG_AUTHOR` environment variable will contain the same value as `VELA_BUILD_SENDER`:

https://github.com/go-vela/types/blob/4dc404f4aabc4ba0f3629113476782f15106085b/library/build.go#L73

Have we considered encouraging usage of `VELA_BUILD_SENDER` instead of `VELA_BUILD_AUTHOR`?

Seems like that would enable the use-case w/o adding another environment variable.

In case my line of questioning seems funny, I'm trying to understand the value of `VELA_BUILD_TAG_AUTHOR` here?

ecrupper: Yes that would seem to be the case. An action step moving forward could be to update the docs to note the `tagger` and `sender` connection. I'm fine closing this PR if people agree that should be the way forward -- especially since email doesn't seem to be obtainable.

JordanSussman: >  I'm fine closing this PR if people agree that should be the way forward 

That would be my vote.

> especially since email doesn't seem to be obtainable.

FWIW, we built a process/API outside of GitHub to map the GitHub username to the underlying Active Directory user to handle this issue.

ecrupper:
Going to close this in favor of providing clearer documentation on using `VELA_BUILD_SENDER` as a means of finding the tagger

PR #213, opened 2021-11-15
chore(linter): remove magic number rule

ecrupper:
Made some changes to disable the magic number rule from the golangci linter. While working on attempting to address all the instances of the rule being broken, I noticed that there were several instances of the rule being manually disabled. I would love some feedback and conversation to make sure this would be a change worth making. Thanks!

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -81,7 +81,6 @@ linters:
     - gofmt
     - goimports
     - golint
-    - gomnd
     - goprintffuncname
     - gosec
     - gosimple # enabled by default
@@ -114,6 +113,7 @@ linters:
   # - gocritic
   # - godox
   # - goerr113
+  # - gomnd
   # - interfacer
   # - nestif
   # - noctx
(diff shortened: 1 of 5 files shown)


wass3r (approved):
i like it. this definitely opens up a bigger conversation around the linters in use. i think `gomnd` just hast too many false positives. there are a few others, i would argue, provide little value. in fact, some just create too much noise. i started a discussion here: https://github.com/go-vela/community/discussions/431 for `lll` for example.

i vote that we try to keep the repos in sync though for linters so whatever change is made here should propagate to the other repos. i'd use this repo as the example to tune the linter and then make one PR to the other repos.

kneal:
Also, like it! For everyone else, we had a small discussion on https://github.com/go-vela/types/pull/211 that led to this.

Yeah, there's a number of linters that could be tweaked to have a more ideal output. 

Also, agree this should be propagated if you have the bandwidth.

kneal (approved):
LGTM 🐬

jbrockopp (approved):
LGTM

PR #214, opened 2021-12-06
fix: check for nil service and step

JordanSussman:
closes https://github.com/go-vela/community/issues/450


wass3r (approved):
`invalid service/step with nil content found` is a user facing error message, right? i wonder if we should make it

KellyMerrick (approved):
lgtm

PR #216, opened 2021-12-09
refactor(version): consolidate version package from plugins

GregoryDosh:
Moving this information into the `types` package will allow us to use this across [all of the different Vela plugins that use this file](https://github.com/search?q=org%3Ago-vela+filename%3Aversion.go). I have an example of this working in a plugin I'm writing internally at the moment and have dropped a link in the Slack channel. The biggest change is really just that the compile time flags look like this now
```sh
LD_FLAGS = -X github.com/go-vela/types/version.Commit=${GITHUB_SHA} -X github.com/go-vela/types/version.Date=${BUILD_DATE} -X github.com/go-vela/types/version.Go=${GOLANG_VERSION} -X github.com/go-vela/types/version.Tag=${GITHUB_TAG}

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/go-vela/types
 go 1.16
 
 require (
+	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/buildkite/yaml v0.0.0-20181016232759-0caa5f0796e3
 	github.com/drone/envsubst v1.0.3
 	github.com/ghodss/yaml v1.0.0
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,5 @@
+github.com/Masterminds/semver/v3 v3.1.1 h1:hLg3sBzpNErnxhQtUy/mmLR2I9foDujNK030IGemrRc=
+github.com/Masterminds/semver/v3 v3.1.1/go.mod h1:VPu/7SZ7ePZ3QOrcuXROw5FAcLl4a0cBrbBpGY/8hQs=
 github.com/aymerick/douceur v0.2.0 h1:Mv+mAeH1Q+n9Fr+oyamOlAkUNPWPlA8PPGR0QAaYuPk=
 github.com/aymerick/douceur v0.2.0/go.mod h1:wlT5vV2O3h55X9m7iVYN0TBM0NH/MmbLnd30/FjWUq4=
 github.com/buildkite/yaml v0.0.0-20181016232759-0caa5f0796e3 h1:q+sMKdA6L8LyGVudTkpGoC73h6ak2iWSPFiFo/pFOU8=
--- a/version/version.go
+++ b/version/version.go
@@ -6,6 +6,11 @@ package version
 
 import (
 	"fmt"
+	"runtime"
+
+	"github.com/Masterminds/semver/v3"
+
+	"github.com/sirupsen/logrus"
 )
 
 const versionFormat = `{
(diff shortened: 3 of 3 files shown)


kneal:
Is there any concern about this generic of content being added here? I'm a little concerned about this repo being turned into a dumping ground. 

I really like the idea of consolidating plugin helper functions and interfaces into a single repo. I'm just not convinced this is the right place for that type of content. Any thoughts?

GregoryDosh:
> Is there any concern about this generic of content being added here? I'm a little concerned about this repo being turned into a catch-all.
> 
> I really like the idea of consolidating plugin helper functions and interfaces into a single repo. I'm just not convinced this is the right place for that type of content. Any thoughts?

I had been talking about this a little bit with @wass3r and he suggested pushing this piece into the types package along with creating another helper utility for the open source rewrites of the SCP & SSH plugin internally. They're using a binary wrapper package and I noticed the version.go file being copied and pasted across multiple plugins. I was suggesting a sort of plugins utility type package that would make writing plugins (in Go) a bit easier. It might make sense to toss this into a go-sdk aimed at plugins instead of interacting with the Vela platform itself.

kneal:
yeah, we have a few things that are often copied across projects and I think that's partly because of some Go principles around sometimes it's easy to copier than abstract. 

I do like the idea of consolidation and I think `version.go` is a good candidate for that; I just think if the purpose of this consolidation is for plugins types isn't really the best place for it IMO. Types is specifically for application code, however, it is frequently used in plugins because of the types of structs and functions it contains. 

I would much prefer the path of a new repo for creating utility-style packages for plugins. I'm not personally a fan of mixing the two (plugin and application) and the rework since its package updates would be annoying long term. With that being said, if people want to overrule my suggestion totally okay. I just don't think long-term this is a very good pattern and I'd rather we do it right the first time than create tech debt for ourselves to fix. 

My suggestion would be to create a new `go-vela/plugin-utils` or `go-vela/plugin-pkg` to manage catch-alls. I think those would both couple with a `plugin-template` repo as well for Go since so many of our plugins are written in similarly.

GregoryDosh:
I don't disagree on the intent to place this code elsewhere. I had originally suggested something along the lines of a utility package or some other package to hold plugin specific sdk-go type stuff but it was suggested to put the code in here because of the relationship with version from the types package.

I think the naming should be suggestive of the language the plugin utilities are for, however, since for now most of the plugins are Go based but that doesn't always need to be the case. Just like the sdk for Vela includes a few language variants, there might also end up being multiple plugins written in different languages.

Perhaps an opinionated way of writing plugins for particular languages so that it makes bootstrapping them easier for end users. This is sort of my other goal in rewriting a few plugins for Open Source, is to wrap a lot of the common functionality and make it so that it's a bit simpler to understand what is going on in the plugin and so that others can mimic and write their own fairly easily if wanting.

`go-vela/plugin-sdk-go` perhaps? Something else?

jbrockopp:
@kneal @GregoryDosh I understand why we would put this code here since all plugins already reference this package.

But I like the idea of creating a pattern now for the future as mentioned i.e. a separate repo for plugin based logic 👍 

But I'm curious, do we need to create a separate repo for language specific plugin utilities?

Thinking we could have one `plugin-helpers` repo (or w/e we call it) and currently it would only have Golang based code.

But at some point, I could see us possibly including helper/utility functions for other languages inside that same repo?

The value there being that it provides one, centrally managed location for all plugin utility functions.

Thoughts?

kneal:
Alternatively, if we wanted repos by language we could always just rename the repository. I don't think we need to be super hard set on that. Right now, I think Go plugins are really our main design. 

I don't really like using the `sdk` naming in this sense because I think we should try and keep things distinct to lower the potential of confusion. I would vote for `helpers` or `utils`. I'm perfectly happy syncing up to and creating the a new repository if that's the biggest roadblock.

GregoryDosh:
I'm also going to close this PR until we discuss more on how this might look going forward from the Slack channel conversation.

wass3r:
i know i'm late to the party here. the thread in slack seems to deal mostly with the binary wrapper and other utils, so my commentary is here, focused on the "version" aspect.

i get the general concern raised by @kneal but i think version is a little special. it is being used by at least three core repos (not plugins) so there would already be utility there.

i don't see the harm for plugins to pull it from types either. most plugins (at least vela maintained ones) are already pulling in the `types` package.

as for the binary wrapper or utility package aspect, i was also in agreement for that to be separate.

PR #217, opened 2021-12-10
enhance(webhook): enable skipping builds during tag events

JordanSussman:
xref: https://github.com/go-vela/community/issues/454

### DIFF
--- a/webhook.go
+++ b/webhook.go
@@ -31,8 +31,8 @@ type Webhook struct {
 // associated with the given hook to determine
 // whether the hook should be skipped.
 func (w *Webhook) ShouldSkip() (bool, string) {
-	// push event
-	if strings.EqualFold(constants.EventPush, w.Build.GetEvent()) {
+	// push or tag event
+	if strings.EqualFold(constants.EventPush, w.Build.GetEvent()) || strings.EqualFold(constants.EventTag, w.Build.GetEvent()) {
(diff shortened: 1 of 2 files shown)


jbrockopp (approved):
LGTM

webhook.go
```diff
@@ -31,8 +31,9 @@ type Webhook struct {
 // associated with the given hook to determine
 // whether the hook should be skipped.
 func (w *Webhook) ShouldSkip() (bool, string) {
-	// push event
-	if strings.EqualFold(constants.EventPush, w.Build.GetEvent()) {
+	// push or tag event
+	// nolint: lll // ignore long line length due to if logic
+	if strings.EqualFold(constants.EventPush, w.Build.GetEvent()) || strings.EqualFold(constants.EventTag, w.Build.GetEvent()) {
```
jbrockopp, line 36: This change LGTM so I already approved 👍 

However, wanted to throw out a question as a discussion for future reviewers.

Is there an appetite to remove the event parsing logic completely?

This would expose the skip functionality to all event types rather than just `push` and `tag`.

wass3r: i think that's fine?! i guess it's only there to lock conditions to _known_ scenarios. for example, i don't know if the commit is null for other events that we want to process, ie. comments? we're probably good, but that would be going through my head.

wass3r (approved):

PR #218, opened 2021-12-14
feat(repo): Add Name History field to Repo type

ecrupper:
[Related Issue](https://github.com/go-vela/community/issues/457)

Adding the Name History field to the repo object. This would be represented as an array of names. While I can't think of any immediate ideas for keeping a history log like this, I figured it would be better than just storing a single previous name. If there is context needed for how this would be used, I have a [Draft PR](https://github.com/go-vela/server/pull/556) in the server where I implement this.


constants/limit.go
```diff
@@ -17,4 +17,7 @@ const (
 
 	// FavoritesMaxSize defines the maximum size in characters for user favorites.
 	FavoritesMaxSize = 5000
+
+	// RepoNameHistoryMax defines the maximum size in characters for repo name history.
+	RepoNameHistoryMax = 5000
```
jbrockopp, line 22: Could you elaborate on why we chose `5000` here?

I'm wondering if this value should be smaller since it seems a bit large.

Based off the PR description, it doesn't sound like we have a use-case for needing a long history.

This has me leaning towards a `n-1` approach for the number of old repo names we keep on record.

library/repo.go
```diff
@@ -348,6 +349,14 @@ func (r *Repo) GetPipelineType() string {
 	return *r.PipelineType
 }
 
+func (r *Repo) GetNameHistory() []string {
+	if r == nil || r.NameHistory == nil {
+		return []string{}
+	}
+
+	return *r.NameHistory
+}
```
jbrockopp, line 358: Could we add comments to this function like the other `Getters` have?

library/repo.go
```diff
@@ -621,6 +630,15 @@ func (r *Repo) SetPipelineType(v string) {
...
 }
 
+func (r *Repo) SetNameHistory(v []string) {
+	// return if Repo type is nil
+	if r == nil {
+		return
+	}
+
+	r.NameHistory = &v
+}
```
jbrockopp, line 640: Could we add comments to this function like the other `Setters` have?

jbrockopp (changes requested):

constants/limit.go
ecrupper: Do you think we would be better served just having the `NameHistory` be just `PreviousName` and have it just be a string type? Perhaps a character limit of 100 or so might be a better idea if we wanted to keep NameHistory. Also, thinking more about this, we would need some sort of clear space logic (rather than rejection logic seen in Favorites) if we didn't want to run into any issues down the line.

ecrupper: I decided to go ahead and give the slice a limit of 100. I also added some truncation logic in the setter method in order to keep an up-to-date name history without erroring out or running over the max. Let me know what you think!

jbrockopp: > Do you think we would be better served just having the NameHistory be PreviousName and have it be a string type?

That would be my preference since we don't have a use case for needing a longer history.

If I'm following correctly, any names older than `n-1` would be considered stale.

If there is a need to have a longer history than that, I would rather defer to the SCM systems audit/history.

All that being said, perhaps we should wait to hear feedback/thoughts from others?

Unless we want me to be the sole individual responsible for making this decision 😅 

> ... we would need some sort of clear space logic (rather than rejection logic seen in Favorites) when the limit is reached if we didn't want to run into any issues down the line.

This debatably could be another reason to only preserve the last name for a repo rather than multiple.

I could see a scenario for confusion if one member of a team renames a repo and a different member was unaware of this change and now they're confused what happened to their repo and why it is no longer in their favorites.

ecrupper: The only thing I can think of is if we wanted to do some database clean up with repos that had no related SCM instance, we could use the NameHistory to find those.  Other than that though, I cannot see a possible reason for holding on to them.

jbrockopp: If I'm following correctly, I think that use case could still be accomplished with `n-1` repo name history 👍 

It sounds like the idea would be to remove any DB records from the `repos` table that don't match:

* the `repo.Name` field for any existing repo
* anything inside the `repo.NameHistory` field

If so, then I think this would also work for checking `repo.PreviousName` instead of checking `repo.NameHistory`.

ecrupper: I apologize -- I misinterpreted my own code. Since this does an `update`, the `NameHistory` field wouldn't have any connection to repos that are in the database. So I think having `repo.PreviousName` be a string value is the way to go.

jbrockopp: No worries 😃 

It's all good 👍 

FWIW:

You could potentially wait to see if others have thoughts/feedback on the discussion before making a change.

wass3r: i like the thought here 👍🏼 

I see the length was changed to 100. i think GitHub's limit is 100. However, Gitlab has a limit of 255 if i'm not mistaken. Wonder if we should take inventory quick of max length from, at least, major SCM providers to see what makes the most sense there.

jbrockopp: I think `RepoNameHistoryMax` could be removed if we decide to only store one previous repo name as a `string` 👍

ecrupper: I just pushed up what the code would look like if we went with just a `string` type.  Can easily be reverted back if we decide to do a slice.

database/repo.go
```diff
@@ -36,6 +36,10 @@ var (
 	// ErrEmptyRepoVisibility defines the error type when a
 	// Repo type has an empty Visibility field provided.
 	ErrEmptyRepoVisibility = errors.New("empty repo visibility provided")
+
+	// ErrExceededPreviousNameLimit defines the error type when a
+	// Repo type has exceeded the limit for the PreviousName field.
+	ErrExceededPreviousNameLimit = errors.New("name history exceeded limit")
```
jbrockopp, line 42: I think we can remove this as it is no longer needed 👍

database/repo.go
```diff
@@ -61,6 +65,7 @@ type Repo struct {
 	AllowTag     sql.NullBool   `sql:"allow_tag"`
 	AllowComment sql.NullBool   `sql:"allow_comment"`
 	PipelineType sql.NullString `sql:"pipeline_type"`
+	PreviousName sql.NullString `sql:"name_history"`
```
jbrockopp, line 68: Should this be `previous_name`?

```suggestion
	PreviousName sql.NullString `sql:"previous_name"`
```

library/repo.go
```diff
@@ -31,6 +33,7 @@ type Repo struct {
 	AllowTag     *bool   `json:"allow_tag,omitempty"`
 	AllowComment *bool   `json:"allow_comment,omitempty"`
 	PipelineType *string `json:"pipeline_type,omitempty"`
+	PreviousName *string `json:"name_history,omitempty"`
```
jbrockopp, line 36: Should this be `previous_name`?

```suggestion
	PreviousName *string `json:"previous_name,omitempty"`
```

library/repo.go
```diff
@@ -348,6 +351,19 @@ func (r *Repo) GetPipelineType() string {
 	return *r.PipelineType
 }
 
+// GetPreviousName returns the PreviousName field.
+//
+// When the provided Repo type is nil, or the field within
+// the type is nil, it returns the zero value for the field.
+func (r *Repo) GetPreviousName() string {
+	// return zero value if Repo type or NameHsitory field is nil
```
jbrockopp, line 359: Should this be `PreviousName`?

```suggestion
	// return zero value if Repo type or PreviousName field is nil
```

library/repo.go
```diff
@@ -621,6 +637,14 @@ func (r *Repo) SetPipelineType(v string) {
 	r.PipelineType = &v
 }
 
+func (r *Repo) SetPreviousName(v string) {
```
jbrockopp, line 640: Could we add comments here similar to the other `SetField()` functions?

library/repo.go
```diff
@@ -621,6 +637,14 @@ func (r *Repo) SetPipelineType(v string) {
...
+	// return if Repo type is nil
+	if r == nil {
+		return
+	}
+	r.PreviousName = &v
```
jbrockopp, line 645: I think we're missing a space here:

```suggestion
	}

	r.PreviousName = &v
```

jbrockopp (changes requested):

library/repo.go
ecrupper: When you do a find and replace routine but misspell `NameHistory`, I suppose things like this happen 🙃

jbrockopp (approved):
LGTM

wass3r (approved):

PR #219, opened 2021-12-20
feat(repo): add build limit field

jbrockopp:
Part of the effort for https://github.com/go-vela/community/issues/27 and https://github.com/go-vela/community/issues/162

Based off of https://github.com/go-vela/community/blob/master/proposals/2020/12-16_rate-limiting.md

Related to https://github.com/go-vela/types/tree/matt/rate-limiting

This adds a `BuildLimit` field to the `github.com/go-vela/database.Repo{}` struct:

https://github.com/go-vela/types/blob/cdbc38e6cc36075fe9c2f2cb1707fdd5e7f2fc71/database/repo.go#L52

This adds a `BuildLimit` field to the `github.com/go-vela/library.Repo{}` struct:

https://github.com/go-vela/types/blob/cdbc38e6cc36075fe9c2f2cb1707fdd5e7f2fc71/library/repo.go#L24

> NOTE:
>
> I chose `BuildLimit` over `Limit` because I thought it was more descriptive field.
>
> However, it does go against the pattern already established with `Counter` and `Timeout`.

Additionally, added some build limit constants that will be used in [go-vela/server](https://github.com/go-vela/server):

https://github.com/go-vela/types/blob/cdbc38e6cc36075fe9c2f2cb1707fdd5e7f2fc71/constants/limit.go#L9-L16

> NOTE:
>
> Open to discussion on these since I wasn't sure what the default and maximum limit for a repo in Vela should be.


constants/limit.go
```diff
@@ -6,6 +6,15 @@ package constants
...
 // Limits and constraints.
 const (
+	// BuildLimitMin defines the minimum value for repo concurrent build limit.
+	BuildLimitMin = 1
+
+	// BuildLimitMax defines the maximum value for repo concurrent build limit.
+	BuildLimitMax = 30
+
+	// BuildLimitDefault defines the default value for repo concurrent build limit.
+	BuildLimitDefault = 10
```
jbrockopp, line 16: Do people feel these values are good in their current state?

database/repo.go
```diff
@@ -49,6 +49,7 @@ type Repo struct {
 	Link         sql.NullString `sql:"link"`
 	Clone        sql.NullString `sql:"clone"`
 	Branch       sql.NullString `sql:"branch"`
+	BuildLimit   sql.NullInt32  `sql:"build_limit"`
```
jbrockopp, line 52: Should this be `Limit` instead of `BuildLimit`?

I lean towards `BuildLimit` because it feels more descriptive on what the field is for.

However, it does diverge from the pattern established with `Counter` and `Timeout`.

If people like this new pattern, then I propose in a future PR, we update:

* `Counter` -> `BuildCounter`
* `Timeout` -> `BuildTimeout`

ecrupper: I like `BuildLimit` and agree with your idea to implement that pattern for the other fields related to builds

GregoryDosh:
I'm a little confused on the purpose of `BuildLimitMin` can you help me understand the use case or scenario where it would be valuable to change that to a different number? It's probably pretty clear in some context but I'm drawing a blank on how I would use that lever in administering a repo or a Vela server instance.

jbrockopp:
> I'm a little confused on the purpose of `BuildLimitMin` can you help me understand the use case or scenario where it would be valuable to change that to a different number? It's probably pretty clear in some context but I'm drawing a blank on how I would use that lever in administering a repo or a Vela server instance.

Thanks for your review 🎉 

The variable is more or less copied from the existing build timeout logic for repos we have today:

https://github.com/go-vela/types/blob/28210cfa70c925b98173ea2f89798c6863238e83/constants/limit.go#L9-L16

The purpose of the variable is used in the [go-vela/server](https://github.com/go-vela/server) in the API call for [updating a repo](https://github.com/go-vela/server/blob/0863d54c5bd726d80e4655e6d15657882ce6ec29/api/repo.go#L614-L627):

```go
	if input.GetTimeout() > 0 {
		// update build timeout if set
		r.SetTimeout(
			int64(
				util.MaxInt(
					constants.BuildTimeoutMin,
					util.MinInt(
						int(input.GetTimeout()),
						constants.BuildTimeoutMax,
					), // clamp max
				), // clamp min
			),
		)
	}
```

If someone updates a repo with a different timeout, we have to ensure the new value is supported.

We accomplish this by "clamping" the value to a specific range (`1m - 90m`).

For a repo to have a timeout outside of that range, it would require involvement from the Vela platform admins.

This same strategy and pattern is adopted for the build limit functionality I'm adding to Vela 👍

GregoryDosh:
> > I'm a little confused on the purpose of `BuildLimitMin` can you help me understand the use case or scenario where it would be valuable to change that to a different number? It's probably pretty clear in some context but I'm drawing a blank on how I would use that lever in administering a repo or a Vela server instance.
> 
> @GregoryDosh Thanks for your review 🎉
> 
> The variable is more or less copied from the existing build timeout logic for repos we have today:
> 
> https://github.com/go-vela/types/blob/28210cfa70c925b98173ea2f89798c6863238e83/constants/limit.go#L9-L16
> 
> The purpose of the variable is used in the [go-vela/server](https://github.com/go-vela/server) in the API call for [updating a repo](https://github.com/go-vela/server/blob/0863d54c5bd726d80e4655e6d15657882ce6ec29/api/repo.go#L614-L627):
> 
> ```go
> 	if input.GetTimeout() > 0 {
> 		// update build timeout if set
> 		r.SetTimeout(
> 			int64(
> 				util.MaxInt(
> 					constants.BuildTimeoutMin,
> 					util.MinInt(
> 						int(input.GetTimeout()),
> 						constants.BuildTimeoutMax,
> 					), // clamp max
> 				), // clamp min
> 			),
> 		)
> 	}
> ```
> 
> If someone updates a repo with a different timeout, we have to ensure the new value is supported.
> 
> We accomplish this by "clamping" the value to a specific range of `1m` - `90m` (`BuildTimeoutMin` - `BuildTimeoutMax`).
> 
> For a repo to have a timeout outside of that range, it would require involvement from the Vela platform admins.
> 
> This same strategy and pattern is adopted for the repo build limit functionality I'm adding to Vela 👍

That makes a lot of sense. I was thinking that this was used in a different context so that additional context helps.

I was trying to understand what would happen if someone put the `BuildLimitMin` to say 5 and then had 4 running builds and then submitted a 5th one what would happen to all the existing things.

Knowing that it's to clamp between valid values makes more sense than allowing users to configure that Min to something else and what the intended behavior would be for end users for values other than 1.

constants/limit.go
JordanSussman: I'm good with these being the default given the fact that we will allow the administrator to override them of the server if they want.

database/repo.go
JordanSussman: Any reason for choosing int32 over int64? If not, can we use int64 to be more consistent with the other items in the struct?

library/repo_test.go
```diff
@@ -110,6 +111,10 @@ func TestLibrary_Repo_Getters(t *testing.T) {
 			t.Errorf("GetBranch is %v, want %v", test.repo.GetBranch(), test.want.GetBranch())
 		}
 
+		if test.repo.GetBuildLimit() != test.want.GetBuildLimit() {
+			t.Errorf("GetTimeout is %v, want %v", test.repo.GetBuildLimit(), test.want.GetBuildLimit())
```
JordanSussman, line 115:
```suggestion
			t.Errorf("GetBuildLimit is %v, want %v", test.repo.GetBuildLimit(), test.want.GetBuildLimit())
```

JordanSussman (changes requested):
small changes

library/repo_test.go
jbrockopp: fixed 👍

database/repo.go
jbrockopp: @JordanSussman yeah the main reason was that `int32` is smaller than `int64` 👍 

In general, `BuildLimit` should contain a small value since we only allow a range of `1 - 30`.

Even if it was updated beyond `30`, I wouldn't expect the limit to exceed a value that `int32` couldn't support.

That being said, If we like changing it to `int64` for consistency than I can get behind that 😄

JordanSussman: @jbrockopp I would prefer `int64` for consistency.

jbrockopp: @JordanSussman this is complete 👍

JordanSussman (dismissed):

constants/limit.go
```diff
@@ -6,6 +6,15 @@ package constants
 
 // Limits and constraints.
 const (
+	// BuildLimitMin defines the minimum value for repo concurrent build limit.
+	BuildLimitMin = 1
+
+	// BuildLimitMax defines the maximum value for repo concurrent build limit.
+	BuildLimitMax = 30
```
kneal, line 13: How do we feel about making this a ratio of the default or allowing admins to set it?

I know for my installation I'm going to want to up the default limit to be 25 which doesn't give a lot of allowances for users to increase on their own. So, I'd propose we either make this max configurable or make it a ratio so it's always got a proportional increase available for users to increase their limit.

jbrockopp: > How do we feel about making this a ratio of the default or allowing admins to set it?

I'm sure we could make either approach work 😄 

But, if I had to choose I would probably pick allowing admins to configure the max limit (and timeout) for builds.

I think that would be a simpler and more straightforward approach 👍 

> I know for my installation I'm going to want to up the default limit to be 25 which doesn't give a lot of allowances for users to increase on their own

To make sure I'm following properly, sounds like your repos will need to run at least `25` builds concurrently 😃 

If so, do you have any insight into how often you see these repos running `25` or more builds at once?

Like once a day? Few times a day? Hourly?

Trying to get a feel for the activity level of your installation 👍

JordanSussman: > So, I'd propose we either make this max configurable or make it a ratio so it's always got a proportional increase available for users to increase their limit.

I agree that the server should be updated to override the maximum number of builds with a static number and/or ratio.

jbrockopp: Doesn't sound like there are any concerns about allowing a configurable max vs ratios?

For now, I'll start working on enabling this functionality in https://github.com/go-vela/server/pull/558

kneal: > If so, do you have any insight into how often you see these repos running 25 or more builds at once?

> Like once a day? Few times a day? Hourly?

> Trying to get a feel for the activity level of your installation

Hard to say right now because the installation isn't fully up but comparing metrics to another CI in the environment. I imagine I'd have at least 10-15 repos that will be running 50-80 builds an hour and the metrics in the graphs appear bursty. So, sometimes it's a steady chug but others randomly fire large quantities at a time. 

Max configurable sounds great to me!! 😄

library/repo.go
```diff
@@ -682,19 +711,20 @@
...
+		r.GetBuildLimit(),
 		r.GetClone(),
+		r.GetCounter(),
 		r.GetFullName(),
 		r.GetID(),
 		r.GetLink(),
 		r.GetName(),
 		r.GetOrg(),
+		r.GetPipelineType(),
+    r.GetPreviousName(),
```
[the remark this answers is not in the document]
jbrockopp: fixed 👍

library/repo_test.go
```diff
@@ -324,20 +335,21 @@
...
+		r.GetBuildLimit(),
 		r.GetClone(),
+		r.GetCounter(),
 		r.GetFullName(),
 		r.GetID(),
 		r.GetLink(),
 		r.GetName(),
 		r.GetOrg(),
+		r.GetPipelineType(),
+    r.GetPreviousName(),
```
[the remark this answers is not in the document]
jbrockopp: fixed 👍

constants/limit.go
jbrockopp: Thanks for the extra information 😃 

The functionality that was requested has been added to https://github.com/go-vela/server/pull/558

JordanSussman (approved):

kneal (approved):
LGTM 🐬

wass3r (approved):

PR #220, opened 2021-12-21
feat(secrets): adding logging mask constant 

ecrupper:
[Related Issue](https://github.com/go-vela/community/issues/136)

Jenkins and Drone both manage to mask secrets in their logs using the mask `***`. I understand there may be concern of redundancy with the `SecretMask` constant, so I'm open to discussion on whether we should use that instead. However, it may be nice to have that consistency with other CI/CD platforms.

Usage in [this PR](https://github.com/go-vela/worker/pull/254)

### DIFF
--- a/constants/secret.go
+++ b/constants/secret.go
@@ -17,4 +17,7 @@ const (
 
 	// SecretMask defines the secret mask to be used in place of secret values returned to users.
 	SecretMask = "[secure]"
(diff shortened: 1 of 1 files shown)


jbrockopp (approved):
LGTM

davidvader (approved):

PR #222, opened 2021-12-23
feat(library): add Duration() function to resources

jbrockopp:
Currently, we have logic to calculate the duration (amount of time a resource was running) for various resources.

The following resources have this functionality:

* [build](https://github.com/go-vela/cli/blob/master/action/build/table.go#L135-L158)
* [service](https://github.com/go-vela/cli/blob/master/action/service/table.go#L131-L154)
* [step](https://github.com/go-vela/cli/blob/master/action/step/table.go#L131-L154)

This change introduces a `Duration()` method on each of these resources that will replace the above linked functions:

https://github.com/go-vela/types/blob/16db0dbc3136fa4a3731b2dcc6c8f32271d3f5d2/library/build.go#L51-L70

https://github.com/go-vela/types/blob/16db0dbc3136fa4a3731b2dcc6c8f32271d3f5d2/library/service.go#L36-L55

https://github.com/go-vela/types/blob/16db0dbc3136fa4a3731b2dcc6c8f32271d3f5d2/library/step.go#L38-L57

### DIFF
--- a/library/step.go
+++ b/library/step.go
@@ -38,7 +38,7 @@ type Step struct {
 // Duration calculates and returns the total amount of
 // time the step ran for in a human-readable format.
 func (s *Step) Duration() string {
-	// check if the service doesn't have a started or finished timestamp
+	// check if the step doesn't have a started or finished timestamp
 	if s.GetStarted() == 0 || s.GetFinished() == 0 {
 		// return zero value for time.Duration (0s)
 		return new(time.Duration).String()
--- a/library/step.go
+++ b/library/step.go
@@ -38,17 +38,16 @@ type Step struct {
 // Duration calculates and returns the total amount of
 // time the step ran for in a human-readable format.
 func (s *Step) Duration() string {
-	// capture finished unix timestamp from the step
-	finished := time.Unix(s.GetFinished(), 0)
-	// capture started unix timestamp from the step
-	started := time.Unix(s.GetStarted(), 0)
-
 	// check if the service doesn't have a started or finished timestamp
 	if s.GetStarted() == 0 || s.GetFinished() == 0 {
 		// return zero value for time.Duration (0s)
 		return new(time.Duration).String()
 	}
 
+	// capture finished unix timestamp from the step
+	finished := time.Unix(s.GetFinished(), 0)
+	// capture started unix timestamp from the step
+	started := time.Unix(s.GetStarted(), 0)
 	// calculate the duration by subtracting the step
 	// started time from the step finished time
 	duration := finished.Sub(started)
--- a/library/step.go
+++ b/library/step.go
@@ -7,6 +7,7 @@ package library
 import (
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/go-vela/types/constants"
 	"github.com/go-vela/types/pipeline"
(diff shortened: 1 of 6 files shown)


library/step.go
```diff
@@ -34,6 +35,27 @@ type Step struct {
 	Distribution *string `json:"distribution,omitempty"`
 }
 
+// Duration calculates and returns the total amount of
+// time the step ran for in a human-readable format.
+func (s *Step) Duration() string {
+	// check if the service doesn't have a started or finished timestamp
```
ecrupper, line 41: Change to `check if the step doesn't...`

ecrupper (changes requested):
Looks good 👍  , one small typo that I spotted.

I think this is a lot cleaner than how it was previously set up with the duration helper functions. Thanks!

library/step.go
jbrockopp: This should be addressed

JordanSussman (approved):

ecrupper (approved):

wass3r:
an afterthought, should a build in progress (ie. `finished` is not set yet) return `0s`? or the time since `started`?

jbrockopp:
@wass3r initially, I thought the same thing and actually had written code to accomplish that 👍 

However, when I started writing tests, I stumbled upon a reason why it may not be great for that to be the case 😅 

The `started` timestamp we use in our test objects is `1563474078`:

https://github.com/go-vela/types/blob/a10f36c258de315292db58fe26b6da4ea27dbc53/library/build_test.go#L810

https://github.com/go-vela/types/blob/a10f36c258de315292db58fe26b6da4ea27dbc53/library/service_test.go#L356

https://github.com/go-vela/types/blob/a10f36c258de315292db58fe26b6da4ea27dbc53/library/step_test.go#L449

That UNIX timestamp actually refers to a time on `Thursday, July 18, 2019` 👍 

After removing the `finished` timestamp from a test object, the `Duration()` function returned something like `2y5m10d`.

In most scenarios, the functionality you described for durations would work 😃 

The typical case for a resource to have a `started` timestamp but not a `finished` timestamp is due to it still running.

However, it's possible for a build to get stuck/hung meaning it never gets a `finished` timestamp.

I decided to use `0s` if the `started` or `finished` timestamp aren't set to help prevent misleading information.

wass3r:
Isn't it still misleading though? A running build (not hung) will show its duration as 0s if the callee is to use the returned value as is.

I agree that hung builds are more of an outlier than running builds, so I think it's ok if the duration says something abnormally large.

jbrockopp:
> Isn't it still misleading though?

Yes, it is misleading either way 🤷 

Would you rather have these functions return `...` like we do in the CLI?

> I agree that hung builds are more of an outlier than running builds, ...

Agreed 👍 

> so I think it's ok if the duration says something abnormally large.

Unfortunately, I don't agree 😞

wass3r:
I guess we disagree.

I find it much more bothersome for a healthy running build to say it's been running for 0s (when it hasn't) than a rare hung build to tell me it has been running for 6 months. If used in the CLI (for example), I would not use the value as is. I think folks would prefer either the current running time, or `...`.

jbrockopp:
Would you like me to open up a PR that changes the behavior to `...` then?

The only downside I see to that is `...` isn't actually a valid duration 😞 

If someone were to use the `Duration()` functions [with `time.ParseDuration()`](https://pkg.go.dev/time#ParseDuration) it would always return an error:

https://go.dev/play/p/_sIxzHTyc9n

On a positive note, the user would still get `0s` as the return value since the default for `time.Duration{}` is `0s`.

PR #223, opened 2021-12-28
feat: add database and library pipeline type

jbrockopp:
Part of the effort for https://github.com/go-vela/community/issues/460

This adds a new `Pipeline{}` type to the `github.com/go-vela/types/database` package:

https://github.com/go-vela/types/blob/30e22e29dabe2da5ac299570feb153c934d62c90/database/pipeline.go#L40-L58

> NOTE: The `data` field will be compressed before being stored in the database.

Additionally, this adds a new `Pipeline{}` type to the `github.com/go-vela/types/library` package:

https://github.com/go-vela/types/blob/30e22e29dabe2da5ac299570feb153c934d62c90/library/pipeline.go#L11-L32

These will be used in the [go-vela/server](https://github.com/go-vela/server) to create a new `pipelines` table in the database.

Also included in this change is a new `pipeline_id` field being added to the `Build{}` type:

https://github.com/go-vela/types/blob/30e22e29dabe2da5ac299570feb153c934d62c90/database/build.go#L38

https://github.com/go-vela/types/blob/30e22e29dabe2da5ac299570feb153c934d62c90/library/build.go#L22

And a `ToLibrary()` function was added to the `github.com/go-vela/types/yaml.Build{}` type:

https://github.com/go-vela/types/blob/30e22e29dabe2da5ac299570feb153c934d62c90/yaml/build.go#L26-L64


database/pipeline.go
```diff
@@ -0,0 +1,247 @@
...
+)
+
+// Pipeline is the database representation of a pipeline.
+type Pipeline struct {
+	ID        sql.NullInt64  `sql:"id"`
+	RepoID    sql.NullInt64  `sql:"repo_id"`
+	Number    sql.NullInt32  `sql:"number"`
+	Flavor    sql.NullString `sql:"flavor"`
+	Platform  sql.NullString `sql:"platform"`
+	Ref       sql.NullString `sql:"ref"`
```
jbrockopp, line 43: Should this be `ref` or `commit`?

Initially, I went with `ref` because that's what we call it in the [go-vela/server](https://github.com/go-vela/server/blob/master/api/pipeline.go#L425-L426) codebase:

```go
	// capture query parameters
	ref := ctx.DefaultQuery("ref", repo.GetBranch())
```

And then we pass that `ref` to [the `source.ConfigBackoff()` function](https://github.com/go-vela/server/blob/master/api/pipeline.go#L434-L435):

```go
	// send API call to capture the pipeline configuration file
	config, err := scm.FromContext(ctx).ConfigBackoff(user, repo, ref)
```

However, in other parts of the codebase we pass [a commit SHA to that function](https://github.com/go-vela/server/blob/master/api/webhook.go#L376-L377):

```go
	// send API call to capture the pipeline configuration file
	config, err := scm.FromContext(c).ConfigBackoff(u, r, b.GetCommit())
```

Feels like we should be using `commit` instead of `ref` in the existing code for pipelines?

In the `Build{}` type we store both fields so we could do the same for `Pipeline{}` if there is value?

database/pipeline.go
```diff
@@ -0,0 +1,253 @@
...
+)
+
+// Pipeline is the database representation of a pipeline.
+type Pipeline struct {
+	ID              sql.NullInt64  `sql:"id"`
+	RepoID          sql.NullInt64  `sql:"repo_id"`
+	Number          sql.NullInt32  `sql:"number"`
+	Flavor          sql.NullString `sql:"flavor"`
+	Platform        sql.NullString `sql:"platform"`
+	Ref             sql.NullString `sql:"ref"`
```
jbrockopp, line 43: Should this be `ref` or `commit`?

Initially, I went with `ref` because that's what we call it in the [go-vela/server](https://github.com/go-vela/server/blob/master/api/pipeline.go#L425-L426) codebase:

```go
	// capture query parameters
	ref := ctx.DefaultQuery("ref", repo.GetBranch())
```

And then we pass that `ref` to [the `source.ConfigBackoff()` function](https://github.com/go-vela/server/blob/master/api/pipeline.go#L434-L435):

```go
	// send API call to capture the pipeline configuration file
	config, err := scm.FromContext(ctx).ConfigBackoff(user, repo, ref)
```

However, in other parts of the codebase we pass [a commit SHA to that function](https://github.com/go-vela/server/blob/master/api/webhook.go#L376-L377):

```go
	// send API call to capture the pipeline configuration file
	config, err := scm.FromContext(c).ConfigBackoff(u, r, b.GetCommit())
```

Feels like we should be using `commit` instead of `ref` in the existing code for pipelines?

In the `Build{}` type we store both fields so we could do the same for `Pipeline{}` if there is value?

JordanSussman: Will pipelines be created during a `deployment` event?

database/pipeline.go
```diff
@@ -0,0 +1,253 @@
...
+	Type            sql.NullString `sql:"type"`
+	Version         sql.NullString `sql:"version"`
+	ExternalSecrets sql.NullBool   `sql:"external_secrets"`
+	InternalSecrets sql.NullBool   `sql:"internal_secrets"`
```
JordanSussman, line 47: Isn't `native` used elsewhere in the codebase?

database/pipeline.go
jbrockopp: Yes, a `pipeline` will be created in the database for every webhook that gets triggered.

The exception to this is any conditions we already have setup that prevent a `build` from being created today.

i.e. https://go-vela.github.io/docs/usage/skipping_build/

database/pipeline.go
jbrockopp: `native` is the secret driver (a.k.a. `engine`) used for managing internal secrets

https://go-vela.github.io/docs/tour/secrets/

https://go-vela.github.io/docs/reference/yaml/secrets/#the-engine-tag

However, the nomenclature attached to secrets fetched by the `native` driver has been `internal secrets`:

https://go-vela.github.io/docs/usage/examples/secrets_external/

https://go-vela.github.io/docs/usage/examples/secrets_internal/

database/pipeline.go
JordanSussman: I think having both fields could provide some level of value. However, I would lean towards `commit` if we only want to store one of them.

jbrockopp: This is done 👍

yaml/build.go
```diff
@@ -21,3 +22,43 @@ type Build struct {
 	Steps       StepSlice          `yaml:"steps,omitempty"     json:"steps,omitempty" jsonschema:"oneof_required=steps,description=Provide sequential execution instructions.\nReference: https://go-vela.github.io/docs/reference/yaml/steps/"`
 	Templates   TemplateSlice      `yaml:"templates,omitempty" json:"templates,omitempty" jsonschema:"description=Provide the name of templates to expand.\nReference: https://go-vela.github.io/docs/reference/yaml/templates/"`
 }
+
+// ToLibrary converts the Build type to a library Pipeline type.
+func (b *Build) ToLibrary() *library.Pipeline {
```
ecrupper, line 27: Thoughts on potentially renaming this function to be a bit more clear? Obviously being too verbose is not ideal, but I would expect a `yaml.build.ToLibrary()` function to convert a `yaml.build` to a `library.build`, kinda like how `Template` is done: https://github.com/go-vela/types/blob/5a3c46e550ce7b9969d9acfb843cf1982dd14240/yaml/template.go#L53. 

Maybe `ToPipeline()` or `ToLibraryPipeline()`? It's not ideal either way, but I can see its usage in other repos being confusing.

ecrupper (commented):
Everything looks great to me! I had one question about some function naming, but otherwise awesome stuff.

yaml/build.go
jbrockopp: I don't think we can use `ToPipeline()` because those functions already exist today 😞 

* [yaml.Metadata](https://github.com/go-vela/types/blob/master/yaml/metadata.go#L21-L36)
* [yaml.Ruleset](https://github.com/go-vela/types/blob/master/yaml/ruleset.go#L38-L48)
* [yaml.SecretSlice](https://github.com/go-vela/types/blob/master/yaml/secret.go#L45-L64)
* [yaml.ServiceSlice](https://github.com/go-vela/types/blob/master/yaml/service.go#L35-L58)
* [yaml.StageSlice](https://github.com/go-vela/types/blob/master/yaml/stage.go#L31-L50)
* [yaml.StepSlice](https://github.com/go-vela/types/blob/master/yaml/step.go#L42-L69)
* [yaml.UlimitSlice](https://github.com/go-vela/types/blob/master/yaml/ulimit.go#L30-L47)
* [yaml.VolumeSlice](https://github.com/go-vela/types/blob/master/yaml/volume.go#L29-L46)
* [yaml.WorkerSlice](https://github.com/go-vela/types/blob/master/yaml/worker.go#L16-L23)

These functions convert a `github.com/go-vela/types/yaml` type to a `github.com/go-vela/types/pipeline` type.

As for the other suggestion, the only concern I have for `ToLibraryPipeline()` is due to the length of the name.

However, I'll make that change now since I don't have an alternative to suggest 👍 

FWIW:

The naming convention isn't something I was too concerned about here.

I say that because it's not currently possible to convert a `yaml.Build{}` type to a `library.Build{}` type.

And I don't think that is functionality we can or want to add to the product.

jbrockopp: This is complete 👍

ecrupper (dismissed):

JordanSussman (approved):

JordanSussman (approved):

JordanSussman (approved):

ecrupper (approved):

PR #225, opened 2022-01-06
chore/fix(log): moving secret masking to log type

ecrupper:
Moving the secret masking from the worker to be a method within `library.Log`. Also made it much more robust with its pattern matching with some go-regexp fun stuff.

### DIFF
--- a/library/log.go
+++ b/library/log.go
@@ -49,15 +49,37 @@ func (l *Log) MaskData(secrets []string) {
 	// convert data to string
 	strData := string(l.GetData())
 	for _, secret := range secrets {
-		secret = regexp.QuoteMeta(secret)
-		re := regexp.MustCompile((`(\s|^|"|:|'|\.|,)` + secret + `(\s|$|"|:|'|\.|,)`))
+		// escape regexp meta characters if they exist within value of secret
+		sanitizedStr := regexp.QuoteMeta(secret)
+
+		// find matches in logs for secrets surrounded by whitespace, start of string,
+		// quotation, colon, period, comma, and end of string.
+		// this prevents partial masking of secrets that contain a substring of another
+		// secret.
+		re := regexp.MustCompile((`(\s|^|"|:|'|\.|,)` + sanitizedStr + `(\s|$|"|:|'|\.|,)`))
 		matches := re.FindAllString(strData, -1)
+
+		// take all matches and mask them
 		for _, match := range matches {
-			mask := string(match[0]) + constants.SecretLogMask + string(match[len(match)-1])
+			// construct mask
+			mask := ""
+
+			// if secret was logged as the very first or last thing in the log, don't add
+			// boundary mask values.
+			if match[0] != secret[0] {
+				mask += string(match[0])
+			}
+			mask += constants.SecretLogMask
+			if match[len(match)-1] != secret[len(secret)-1] {
+				mask += string(match[len(match)-1])
+			}
+
+			// replace log with new mask
 			strData = strings.Replace(strData, match, mask, 1)
 		}
-		strData = re.ReplaceAllString(strData, constants.SecretLogMask)
 	}
+
+	// update data field to masked logs
 	l.SetData([]byte(strData))
 }
 
--- a/library/log.go
+++ b/library/log.go
@@ -7,7 +7,6 @@ package library
 import (
 	"fmt"
 	"regexp"
-	"strings"
 
 	"github.com/go-vela/types/constants"
 )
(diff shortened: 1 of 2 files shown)


library/log.go
```diff
@@ -36,6 +42,25 @@ func (l *Log) AppendData(data []byte) {
...
+	for _, secret := range secrets {
+		secret = regexp.QuoteMeta(secret)
+		re := regexp.MustCompile((`(\s|^|"|:|'|\.|,)` + secret + `(\s|$|"|:|'|\.|,)`))
+		matches := re.FindAllString(strData, -1)
+		for _, match := range matches {
+			mask := string(match[0]) + constants.SecretLogMask + string(match[len(match)-1])
+			strData = strings.Replace(strData, match, mask, 1)
+		}
+		strData = re.ReplaceAllString(strData, constants.SecretLogMask)
+	}
```
jbrockopp, line 60: Looks like there is a lot going on here 😅 

Could we add comments to better illustrate what this code is doing?

jbrockopp (changes requested):

library/log.go
```diff
@@ -36,6 +42,47 @@ func (l *Log) AppendData(data []byte) {
...
+			}
+			mask += constants.SecretLogMask
+			if match[len(match)-1] != secret[len(secret)-1] {
+				mask += string(match[len(match)-1])
+			}
+
+			// replace log with new mask
+			strData = strings.Replace(strData, match, mask, 1)
+		}
+	}
```
ecrupper, line 80: To explain the complicated masking procedure, I am trying to address all the following scenarios:

Say your secret was the value `(vela**secret[xyz]{`

`QuoteMeta` will turn that value into `\(vela\*\*secret\[xyz\]\{` in order to find its occurrence in the logs. 

However, if we had another secret, say `xyz`, then there is a potential that the above secret would get masked to `(vela**secret[***]{`. This is why there needed to be the introduction of boundary characters, which are all the characters specified on line 59. It's not foolproof. For example, if we added the characters `[` and `]` to the mix, then the `xyz` secret would get skipped. Hopefully folks can chime in on how best to approach these complicated arrangements of secrets within logs. 

For now, it will simply address scenarios like these:

`"$ echo $SECRET\n(vela**secret[xyz]{\n` --> `"$ echo $SECRET\n***\n`
`(vela**secret[xyz]{\n/bin/sh: (vela**secret[xyz]{: not found` --> `***\n/bin/sh: ***: not found`

And so on. The various ways around the current state could be `<secretValue>` or `[secretValue]`, but that can be fixed by adding those to the pattern we're matching.

library/log.go
```diff
@@ -36,6 +42,47 @@ func (l *Log) AppendData(data []byte) {
 	l.SetData(append(l.GetData(), data...))
 }
 
+// MaskData reads through the log data and masks
+// all values provided in the string slice. If the
+// log is empty, we do nothing.
+func (l *Log) MaskData(secrets []string) {
```
jbrockopp, line 48: Thanks for your work on this 🎉 

My first piece of feedback would be to remove most, if not all, `[]byte` -> `string` conversions.

Unless they are required, this should increase the performance by removing those conversions.

i.e. 

* `strData := string(l.GetData())` -> `data := l.GetData()`
* `matches := re.FindAllString(strData, -1)` -> `matches := re.FindAll(data, -1)`
* `strData = strings.Replace(strData, match, mask, 1)` -> `data = bytes.Replace(data, match, mask, 1)`

However, I'm not an expert with regexp but I wonder if we need to construct a new mask for each match?

Feels like this could be accomplished with the regexp library without extra logic.

Here is a sample of what I mean:

https://go.dev/play/p/vm4IAGT-IUp

A rough example of how this might look in our code:

```go
	data := l.GetData()

	for _, secret := range secrets {
		// escape all regexp metacharacters that exist in the secret
		//
		// https://pkg.go.dev/regexp#QuoteMeta
		escaped := regexp.QuoteMeta(secret)

		// create regexp to match secrets in the log data surrounded by regexp metacharacters
		//
		// https://pkg.go.dev/regexp#MustCompile
		re := regexp.MustCompile(`(\s|^|"|:|'|\.|,)` + escaped + `(\s|$|"|:|'|\.|,)`)

		// create a mask for the secret
		mask := fmt.Sprintf("$1%s$2", constants.SecretLogMask)

		// replace all regexp matches of secret with mask
		//
		// https://pkg.go.dev/regexp#Regexp.ReplaceAll
		data = re.ReplaceAll(data, []byte(mask))

		// update log data with any secret values masked
		l.SetData(data)
	}
```

jbrockopp (changes requested):

library/log.go
ecrupper: Yes -- this seems to work and is much cleaner. Thanks!

jbrockopp (approved):
LGTM

dtanner (approved):

PR #229, opened 2022-01-24
Revert "fix(secret)!: remove implicit repo secrets and clean up secrets testing"

KellyMerrick:
Reverts go-vela/types#210

PR #230, opened 2022-01-24
fix(secret)!: remove implicit repo secrets and clean up secrets testing

KellyMerrick:
This pr is to create a net-new pr for https://github.com/go-vela/types/pull/210, which was reverted in https://github.com/go-vela/types/pull/229. That pr was for a breaking change, that we decided not to introduce into an RC to be cut today.


jbrockopp (changes requested):
> This pr is to create a net-new pr for #210, which was reverted in #229. That pr was for a breaking change, that we decided not to introduce into an RC to be cut today.

Could you elaborate on why that PR was reverted?

I think updating the description for https://github.com/go-vela/types/pull/229 with more details would be ideal

wass3r:
It was an accidental merge. Just prior to it we decided to freeze current state across core repos for the RC unless they were dependency upgrades.

constants/secret.go
```diff
@@ -20,4 +20,7 @@ const (
 
 	// SecretLogMask defines the secret mask to be used when distributing logs that contain secrets.
 	SecretLogMask = "***"
+
+	// SecretRestrictedCharacters defines the set of characters that a secret name cannot contain.
```
kneal, line 24: Can we note what this character is in the human form? It's not clear if you don't know what the character means. Google is always helpful but it would be nice to just state it human friendly.

wass3r: fair callout

jbrockopp (dismissed):
LGTM

constants/secret.go
wass3r: @GregoryDosh do you want to add a commit to this PR with that info? i think it would be worthwhile to call out what `\x00` is

wass3r:
@KellyMerrick anything else needed here?

PR #231, opened 2022-01-31
fix(library): add error field truncating

ecrupper:
Fixes [Issue 468](https://github.com/go-vela/community/issues/468)

On rare occasions, the error generated by a step exceeds our database limit of 500 characters. When this happens, the build hangs and does not complete. This PR makes sure we don't set the error field to a value we can't support in the database.


cognifloyd (commented):
Which part of the error is most likely to include the useful bits? In python, I always want the end of the traceback. I'm still too new to go to say for sure.

What if you took the first and last 250 characters, dropping everything in the middle?

ecrupper:
> Which part of the error is most likely to include the useful bits? In python, I always want the end of the traceback. I'm still too new to go to say for sure.
> 
> What if you took the first and last 250 characters, dropping everything in the middle?

I like this idea. Most errors contain the vital information in the beginning or end.

kneal:
Also, wonder if in addition to character limits if we should just support more characters in for the field in the database. I don't think it would be super harmful to allow 1000 characters.

GregoryDosh:
Is it possible to place the truncation code to the "last responsible moment" so that within any libraries or other bits of code that want to use the full error, they can, but when we go to place the data into the database we're only saving the important bits? I don't have context into if you've already explored that, but could reduce the code duplication and not preemptively truncate before we need to do so.

Otherwise it looks good and I don't see any obvious issues. Thanks for your contribution!

constants/limit.go
```diff
@@ -24,6 +24,9 @@ const (
 	// BuildTimeoutDefault defines the default value in minutes for repo build timeout.
 	BuildTimeoutDefault = 30
 
+	// ErrorLimit defines the maximum size in characters for resource error fields.
```
GregoryDosh, line 27: Oh, one small thing, this comment is helpful, but it doesn't let anyone know what happens if the limit is exceeded. They'd have to dig into the code to understand that half of the beginning and half of the end are kept. It might be beneficial for future folks looking in the codebase to understand the intent of that and what happens if it gets exceeded.

ecrupper:
I relocated these changes to be within the `crop` function of the build database object. It seems like that's where it belongs. I decided to leave out `hook`, `step`, and `service` for now. When trying to reproduce the error from the issue with just the `build` crop updated, it seemed to work just fine.

GregoryDosh (approved):
Nice catch! This all looks good to me.

database/build.go
```diff
@@ -27,6 +27,8 @@ const (
 	maxTitleLength = 1000
 	// Maximum message field length.
 	maxMessageLength = 2000
+	// Maximum error field length.
+	maxErrorLength = 500
```
jbrockopp, line 31: I agree with @kneal 👍 

We should explore expanding the number of chars we allow for the `error` field across all resources.

Doubling it from `500` -> `1000` sounds like a fair place to start 😃

database/build.go
```diff
@@ -75,6 +77,13 @@ func (b *Build) Crop() *Build {
 		b.Message = sql.NullString{String: b.Message.String[:maxMessageLength], Valid: true}
 	}
 
+	// trim the Error field to 500 characters — 250 from each end to maximize usefulness.
+	if len(b.Error.String) > maxErrorLength {
+		front := maxErrorLength - (maxErrorLength / 2)
+		end := len(b.Error.String) - (maxErrorLength / 2)
+		str := b.Error.String[:front] + b.Error.String[end:]
+		b.Error = sql.NullString{String: str, Valid: true}
+	}
```
jbrockopp, line 86: Are we sure we're taking the right approach here?

Personally, I'd rather start with increasing the number of characters before going down this route.

If we feel that's insufficient, then we should use compression for the `error` field.

jbrockopp (changes requested):

database/build.go
wass3r: it's currently a somewhat rare occurrence (afaik) so handling the outliers this way instead of having the outliers shape the data size seems fine to me. do we expect this be more commonplace?

database/build.go
ecrupper: I understand not wanting to process string logic like this, but I am curious as to why the `message` and `title` fields get cropped in a similar fashion? 

Also, I do wonder if it may be more efficient to handle larger errors by referencing the log from which it came rather than attempt to store some truncated version.

database/build.go
kneal: My personal view would be I don't think that field has ever changed in size. So, maybe this is showing errors that could be growing in size. I also don't think `1000` is that big of a concern and then would save us having to parse errors and hopefully not cut off the important part. If we had to go beyond `1000` then I think truncating could be useful.

I'm also not sure if we use error wrapping everywhere in the worker/runtime to make this errors longer so depending how errors are handled or want to be handled long term in that codebase `1000` might be the norm but I honestly don't know. The `error:` field in the build resource is entirely for admins because thats where we populate a platform error. So, cutting off the data really only hurts ourselves for richer research. 

> We should explore expanding the number of chars we allow for the error field across all resources.

@jbrockopp probably not a bad idea since I don't think that's ever been done. I assume most of these char lengths are there original defaults.

ecrupper: So if we bump the limit to 1000, do we still want to crop the value if it exceeds that limit? Or do we have another idea for an approach to that situation?

kneal: The only other idea I have would be compression if we wanna keep the field small but keep the full error messages.

ecrupper: Would you support compressing the other fields currently being processed by the `crop` method? (`message` and `title`)

kneal: I don't know the answer to that one. I think for error it might make sense to crop because that field could be quite variable depending on how the error wrapping works. 

Title and Message might make sense to crop because those come from the SCM system attached and I would assume GitHub has limits on those. So, it might make sense to just mirror those limits.

jbrockopp: @ecrupper 

> So if we bump the limit to 1000, do we still want to crop the value if it exceeds that limit? Or do we have another idea for an approach to that situation?

I'm onboard with the idea of a "simple" cropping strategy similar to `title` and `message`.

In the future, if this becomes a problem with the `1000` char limit, then we introduce compression.

cc: @kneal @wass3r

jbrockopp (approved):
LGTM

wass3r (approved):

PR #236, opened 2022-02-25
feat(rulesets): action constants + unmarshal logic for legacy PR event

ecrupper:
Part of the effort of https://github.com/go-vela/community/pull/502

1) Add new action constants that will be used to form the event type for scoped PR events (e.g. `pull_request:synchronized`)

2) Add logic in the unmarshalling of the `ruleset` to preserve legacy method of `event: [pull_request]`. I decided to keep it as it works today rather than have it default to listen to _all_ actions.

3) Trim scoped events to their parent events when dealing with matching secrets


KellyMerrick (approved):
lgtm

yaml/ruleset.go
```diff
@@ -137,12 +137,24 @@ func (r *Rules) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 		r.Repo = rules.Repo
 		r.Status = rules.Status
 		r.Tag = rules.Tag
 		r.Target = rules.Target
+
+		// account for users who use non-scoped pull_request event
+		events := []string{}
+
+		for _, e := range rules.Event {
+			if e == "pull_request" {
```
JordanSussman, line 150: I prefer to utilize [strings.EqualFold](https://pkg.go.dev/strings#EqualFold) for comparison. 

```suggestion
			if strings.EqualFold(e, "pull_request")  {
```

yaml/ruleset.go
```diff
@@ -137,12 +137,24 @@ func (r *Rules) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+				events = append(events, "pull_request:opened", "pull_request:synchronized")
```
JordanSussman, line 151: Couple of questions:

- Should you utilize the constants defined above?
- Why is `edited` not included?

constants/action.go
```diff
@@ -0,0 +1,17 @@
...
+// Build and repo events.
+const (
+	// ActionOpened defines the action for opening pull requests.
+	ActionOpened = "opened"
+
+	// ActionEdited defines the action for the editing of pull requests.
+	ActionEdited = "edited"
+
+	// ActionSynchronized defines the action for the synchronizing of pull requests.
+	ActionSynchronized = "synchronized"
```
JordanSussman, line 16: Does GitHub generate a webhook when the pull request is closed? If so, can we include that too?

JordanSussman (commented):
initial feedback

yaml/ruleset.go
ecrupper: Sounds good! I'll change that

yaml/ruleset.go
ecrupper: Yeah using the constants is probably a better idea.

As for the absence of `edited`, I wanted to make sure that the `pull_request` event wouldn't change behavior if it wasn't changed. Since we currently only react to `opened` and `synchronize` actions, adding `edited` would result in unexpected behavior. Although, the overall scoping pattern would be much better sensical if `pull_request` applied to all actions. Open for more discussion on that front

JordanSussman: I think we would want to do something like `pull_request:*` or `pull_request:all` to allow backwards compatibility.

yaml/ruleset.go
```diff
@@ -137,12 +137,34 @@ func (r *Rules) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+			case constants.EventPull:
+				events = append(events,
+					constants.EventPull+":"+constants.ActionOpened,
+					constants.EventPull+":"+constants.ActionSynchronized)
+			// convert all and * to cover all configured actions
+			case constants.EventPull + ":" + "all", constants.EventPull + ":" + "*":
+				events = append(events,
+					constants.EventPull+":"+constants.ActionOpened,
+					constants.EventPull+":"+constants.ActionSynchronized,
+					constants.EventPull+":"+constants.ActionEdited)
```
cognifloyd, line 161: I'm not sure about introducing `pull_request:all` without including all of the possible github events. Adding more event types to this later will be another backwards compat issue because the definition of `:all` and `:*` will change. Can we include all known event types now instead of adding them piecemeal?

yaml/ruleset.go
```diff
@@ -137,12 +137,34 @@ func (r *Rules) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		// account for users who use non-scoped pull_request event
+		events := []string{}
+
+		for _, e := range rules.Event {
+			switch e {
+			// backwards compatibility - pull_request = pull_request:opened + pull_request:synchronized
+			case constants.EventPull:
+				events = append(events,
+					constants.EventPull+":"+constants.ActionOpened,
+					constants.EventPull+":"+constants.ActionSynchronized)
```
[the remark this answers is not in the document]
ecrupper: All the instances where the `want` value of the events are `pull_request:opened, pull_request:synchronized` are testing the backwards compatibility.

yaml/ruleset.go
wass3r: i think the argument is that when the definition of "all" or "*" changes (by adding more events/actions whether that's likely or not), all existing pipelines using that would suddenly behave differently.

as implemented in this PR, we achieve backwards compatibility when the user just supplies `pull_request` as they did previously which is great. we might consider just dropping the "all"/"*" idea and only allow to explicitly list each action. arguably, the likelihood of folks actually wanting things to happen on _all_ events is low, and if they did, they still could.

wass3r (approved):
I expect that will want to have a docs PR that talks about this functionality.

yaml/ruleset.go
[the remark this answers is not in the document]
cognifloyd: Oh. Right. I missed that that was in `want`. Cool!

cognifloyd (approved):

PR #239, opened 2022-03-03
feat(hook): add hook id to hook object

ecrupper:
Couple related issues: https://github.com/go-vela/community/issues/459 and https://github.com/go-vela/community/issues/416

The `hook_id` is the field needed from GitHub to not only reach that deliveries page but also interact with any of their API endpoints related to hook deliveries. A `hook_id` is unique to any one hook within a repo, but since we already have an `id` field as well as a `repo_id` field, I chose `address` as the name. If anyone has a suggestion for another name, I am all ears! 

I also cleaned up a couple of typos I noticed in the comments of `build.go`

### DIFF
--- a/library/hook.go
+++ b/library/hook.go
@@ -24,6 +24,7 @@ type Hook struct {
 	Error    *string `json:"error,omitempty"`
 	Status   *string `json:"status,omitempty"`
 	Link     *string `json:"link,omitempty"`
+	Address  *int64  `json:"address,omitempty"`
 }
 
 // GetID returns the ID field.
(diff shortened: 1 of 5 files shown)


JordanSussman:
> I chose address as the name. If anyone has a suggestion for another name, I am all ears!

I'm thinking `hook_id` would be less confusing compared to `address`.

library/hook.go
```diff
@@ -182,6 +183,19 @@ func (h *Hook) GetLink() string {
 	return *h.Link
 }
 
+// GetAddress returns the Address field.
+//
+// When the provided Hook type is nil, or the field within
+// the type is nil, it returns the zero value for the field.
+func (h *Hook) GetAddress() int64 {
+	// return zero value if Hook type or Address field is nil
+	if h == nil || h.ID == nil {
```
plyr4, line 192: typo?
```suggestion
	if h == nil || h.Address == nil {
```

plyr4:
although i see the confusing `hook.GetHookID` and it not returning the ID of the hook to be confusing.

plyr4:
some options, not all of them _good_
- hook_id
- delivery_id
- scm_id
-  webhook_id
- payload_id
- address_id

plyr4 (approved):

jbrockopp (approved):
LGTM

PR #251, opened 2022-04-26
feat(rulesets): add event action field

ecrupper:
I was noticing an issue with [my first iteration on this](https://github.com/go-vela/types/pull/236). In order to execute a step, it must pass [the purge process](https://github.com/go-vela/types/blob/abb9e3afc4bd429783c6805c051e85aa58b93652/pipeline/container.go#L60) and [the skip process](https://github.com/go-vela/worker/blob/151849034f86575046422c1407d33fc1aafa13b1/internal/step/skip.go#L18). In order to respect scoped events (`pull_request:opened` for example), I needed to change the `build.Event` value to that [here](https://github.com/go-vela/server/blob/ccfff2af42eea4abeceac8ac3d5b93790b3ef300/scm/github/webhook.go#L208). This is not ideal. Not only does it require splitting logic all over to respect various event gatekeeping, it also changes the environment variable `VELA_BUILD_EVENT`, which is used by many end users. 

To get around this, I decided to add a new field, `EventAction` (open to new name suggestions). This lays the ground work for future event action handling (such as PR review events). It also makes handling comment event scoping seamless, which I've also added to this PR. 

The obvious drawback to this method is two fold:
1) another column in the database
2) that column will be a duplicate of the `event` column for `push`, `deploy`, and `tag` events. 

Another strategy here would be to change `EventAction` to `Action`, and handle the events without actions as `nil`. I chose the more verbose option for now but am more than willing to pursue that route instead if we wanted.

### DIFF
--- a/constants/action.go
+++ b/constants/action.go
@@ -9,7 +9,10 @@ const (
 	// ActionOpened defines the action for opening pull requests.
 	ActionOpened = "opened"
 
-	// ActionEdited defines the action for the editing of pull requests.
+	// ActionCreated defines the action for creating issue comments.
+	ActionCreated = "created"
+
+	// ActionEdited defines the action for the editing of pull requests or issue comments.
 	ActionEdited = "edited"
 
 	// ActionSynchronized defines the action for the synchronizing of pull requests.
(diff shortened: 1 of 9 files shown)


cognifloyd (approved):
LGTM

ecrupper:
I decided to go with storing just `Action` to avoid redundancy in columns and parity with what GitHub gives us as the `action`

cognifloyd (approved):
Oh I like separating the action like that much better

PR #252, opened 2022-05-09
fix(hook): clean up old repo rename code by adding action field to hook

ecrupper:
Context: there have been some errors related to repository events not being processed correctly by the webhook endpoint. This led me to look over the [renaming a repository code](https://github.com/go-vela/server/pull/556). The fix for the errors can be handled in the server code, but what caught my eye was the fact that I chose to override the `hook.Event` field to be a custom `renameRepository` hard-coded string. With the [new scoped ruleset support](https://github.com/go-vela/server/pull/630), I think it would be consistent to mirror `build.Event` and `build.EventAction` with `hook.Event` and `hook.EventAction`. 

Also random, but VSCode tried to include a project folder in the commit, so I added that to the `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@
 *.ipr
 *.iws
 *.xml
+
+# VSCode project folder
+.vscode/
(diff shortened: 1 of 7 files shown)


cognifloyd (approved):

jbrockopp (approved):
LGTM

wass3r (approved):

PR #253, opened 2022-05-12
fix(log): add = to secret mask

ecrupper:
If a user were to run `printenv`, since we add secrets to the environment, and the printing of the environment is of the style `key=value` with no space, secrets can be printed using that. Further, not including `=` in the pre-secret mask seems like a pretty large oversight.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@
 *.ipr
 *.iws
 *.xml
+
+# VSCode project folder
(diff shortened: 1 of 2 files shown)


wass3r (approved):

cognifloyd (approved):

jbrockopp (approved):
LGTM

PR #255, opened 2022-05-26
feat(worker): add last repo and last build fields

ecrupper:
Part of the effort for the admin page.

I wanted to create a favorable situation for our UI to bring in helpful information to the admin page. Instead of trying to search through our builds table to find out the latest build on a given worker, it's much easier to keep the much smaller worker table up-to-date. 

Further, with [this PR in review](https://github.com/go-vela/server/pull/646), I chose to represent the build in a lightweight fashion: by its ID.

Would love any feedback on this! Issue for context: https://github.com/go-vela/community/issues/610

### DIFF
--- a/database/worker.go
+++ b/database/worker.go
@@ -31,6 +31,8 @@ type Worker struct {
 	Active        sql.NullBool   `sql:"active"`
 	LastCheckedIn sql.NullInt64  `sql:"last_checked_in"`
 	BuildLimit    sql.NullInt64  `sql:"build_limit"`
+	LastRepo      sql.NullString `sql:"last_repo"`
+	LastBuildID   sql.NullInt64  `sql:"last_build_id"`
 }
 
 // Nullify ensures the valid flag for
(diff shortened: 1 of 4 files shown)


jbrockopp (changes requested):
I think my primary concern with this approach is related to thinking into the future.

Perhaps I'm getting ahead of myself and overthinking about this for little value.

> NOTE: I'm leaving this as request changes for visibility purposes.

AFAICT, this path is streamlined for a setup where workers are only running one build.

To be clear, I could certainly get behind that approach using the MVP mindset.

However, what concerns me is what comes after the fact meaning how do we extend support for this to encompass Vela installations that run multiple builds per worker.

Would we modify the existing fields to be slices (arrays)?

e.g. `LastRepos`, `LastBuildIDs` ?

If so, do we think it's worth adding the slices now?

Also, how would the workers know when to remove old repos/builds from those slices?

These are the kinds of things that popped into my head.

I almost wonder if an API driven approach (server -> worker) may be more suitable?

ecrupper:
Closing this for a couple reasons: 1) Jordan brings up good points about future implementation and overall better design options, and 2) this isn't really the direction the admin page is likely to go. Needs rethinking

PR #257, opened 2022-06-08
fix(yaml/ruleset): fix unmarshal PR event to use synchronize, not synchronized

ecrupper:
Since we decided to model off of what GH tells us is the action, I forgot to change this one instance of unmarshaling where I used `synchronized` instead of `synchronize`.


wass3r (approved):

KellyMerrick (approved):
lgtm

PR #266, opened 2022-09-15
fix(k8s/build/sanitize): implement similar line length and lowercase logic as containers for build

ecrupper:
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

Users run into trouble if their org or repo have uppercase letters, or the combination of the two results in a string longer than 63 characters. Fix was already in place for container names, but not pods.

### DIFF
--- a/pipeline/build.go
+++ b/pipeline/build.go
@@ -5,7 +5,9 @@
 package pipeline
 
 import (
+	"fmt"
 	"strings"
+	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


wass3r (approved):
cool with it, but we might consider adding a utility function for sanitizing name according to the spec

JordanSussman (approved):

plyr4 (approved):

PR #277, opened 2023-02-23
feat(worker-visibility): extend worker table with 5 fields

KellyMerrick:
Worker Visibility Proposal: https://github.com/go-vela/community/pull/813

These five fields will be added to the worker table:
- status
- last_status_update_at
- running_build_ids
- last_build_started_at fields
- last_build_finished_at fields

See also: https://github.com/go-vela/server/pull/772 and https://github.com/go-vela/worker/pull/482


library/worker.go
```diff
@@ -86,6 +90,58 @@ func (w *Worker) GetActive() bool {
...
+func (w *Worker) GetLastStatusUpdateAt() int64 {
+	// return zero value if Worker type or LastStatusUpdateAt field is nil
+	if w == nil || w.LastStatusUpdateAt == nil {
+		return 0
+	}
+
+	return *w.LastStatusUpdateAt
+}
+
+// GetRunningBuildIDs returns the Routes field.
```
cognifloyd, line 119:
```suggestion
// GetRunningBuildIDs returns the RunningBuildIDs field.
```

KellyMerrick: Good catch, ty!

wass3rw3rk (approved):
just a question. lgtm tho

database/worker.go
```diff
@@ -100,6 +135,17 @@ func (w *Worker) Validate() error {
...
+	// calculate total size of RunningBuildIds
+	total := 0
+	for _, f := range w.RunningBuildIDs {
+		total += len(f)
+	}
+
+	// verify the RunningBuildIds field is within the database constraints
+	if total > constants.RunningBuildIDsMaxSize {
+		return ErrExceededRunningBuildIDsLimit
+	}
```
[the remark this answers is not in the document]
KellyMerrick: Since there isn't a defined executor limit for a single worker, and the build id lengths grow over time, I took the approach that if build_id's reach into the millions (up to 9 chara) with commas separating the ids (so 10 chara), this limit accommodates 50 builds.

plyr4 (approved):
nice

PR #279, opened 2023-03-01
enhance: Replace magic "init"/"#init" strings with `IsInit` flag on Stage/Step

cognifloyd:
There are a bunch of places where we check for the `"init"` string in the worker. Here's an idea for a flag to replace that magic string.


cognifloyd:
Nope. This doesn't work because the pipeline gets dumped/loaded via yaml when pushed/popped from the queue and because the compile step sends the yaml to the external modify endpoint after the init step has already been injected. This needs more design work.

PR #291, opened 2023-05-17
enhance(repo)!: add AllowSchedule field

jbrockopp:
xref: https://github.com/go-vela/community/issues/538

Part of https://github.com/go-vela/community/pull/772

Based off of https://github.com/go-vela/types/pull/289

This change add an `AllowSchedule` field to the `Repo{}` object:

https://github.com/go-vela/types/blob/eb7c705a753486439d58ad8d460dcaba75d7991d/database/repo.go#L71

https://github.com/go-vela/types/blob/eb7c705a753486439d58ad8d460dcaba75d7991d/library/repo.go#L38

Similar to the other `Allow*` fields, this will enable users to control if their repo responds to the `schedule` event.

Initially, I didn't plan to add this field since the `Allow*` fields control which `git` based events to respond to.

And `schedules` will be totally controlled by Vela (i.e. not SCM driven), and won't be an event for webhooks.

That being said, I wanted to start with opening this PR to collect feedback on if we should add this or not.

You can find an example of how this field will be used within https://github.com/go-vela/server/pull/846.

> NOTE: Marking this as a breaking change since it adds a new field which will require DB migrations.

### DIFF
--- a/database/repo.go
+++ b/database/repo.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 Target Brands, Inc. All rights reserved.
+// Copyright (c) 2023 Target Brands, Inc. All rights reserved.
 //
 // Use of this source code is governed by the LICENSE file in this repository.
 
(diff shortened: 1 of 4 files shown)


ecrupper (approved):

plyr4 (approved):
LGTM, seems like a straightforward way to disable repo schedules without actually removing the schedule or disabling event-driven builds. nice

wass3r (approved):
works for me, but i can see your hesitation about adding non-SCM events here. i think it's fine.

jbrockopp:
@ecrupper @plyr4 @wass3r only one last caveat to point out and then I think we can merge 👍 

The server scan's the entire `schedules` table for all schedules with `active == true`.

This would include any `schedules` for a repo that have the `schedule` event disabled.

This could mean the server spends unneeded extra cycles processing `schedules` for a repo with that event disabled.

We could perform a `JOIN` with the `repos` table to check for the event being disabled?

However, in general, I try to avoid `JOIN` queries since they can be taxing on the system especially for large tables.

cognifloyd:
I think the per repo flag complicates things. It might be a good idea, but the global allow list is sufficient for now, so let's wait till after the rest of schedules are implemented and then work on adding the per-repo flag.

wass3rw3rk:
@jbrockopp thanks for calling attention to that. i'm good with the approach @cognifloyd mentioned.

jbrockopp:
@wass3rw3rk perhaps I misunderstood, but if we want to go with that approach we shouldn't merge this, right?

So I think we'll need to revert this PR and then I can open a new one with this change as a draft PR.

Otherwise, we'll be unable to pull this repo on the `main` branch into [go-vela/server](https://github.com/go-vela/server).

PR #292, opened 2023-05-18
feat!: protect against stale queue items

cognifloyd:
Part of https://github.com/go-vela/community/issues/811
Blocks: https://github.com/go-vela/worker/pull/478

This adds an `ItemVersion` field to `Item` with an `ItemVersion` constant  that should be bumped whenever we make significant changes to `Item` or the `pipeline.Build`. This way, the worker can detect any stale builds that were queued before a Vela server upgrade or downgrade, and fail the build.

In the future, we could do something even nicer and ask the server to recompile and requeue the build.

### DIFF
--- a/item.go
+++ b/item.go
@@ -22,7 +22,7 @@ type Item struct {
 	Repo     *library.Repo   `json:"repo"`
 	User     *library.User   `json:"user"`
 	// The 0-value is the implicit ItemVersion for queued Items that pre-date adding the field.
-	ItemVersion uint64 `json:"version"`
+	ItemVersion uint64 `json:"item_version"`
 }
 
 // ToItem creates a queue item from a pipeline, build, repo and user.
--- a/item.go
+++ b/item.go
@@ -9,20 +9,29 @@ import (
 	"github.com/go-vela/types/pipeline"
 )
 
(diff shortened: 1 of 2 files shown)


plyr4 (approved):
LGTM, that will make future updates a lot smoother.
should we mark this as breaking?

cognifloyd:
It's difficult to say if it is breaking or not. I added the `!`. :shrug:

item.go
```diff
@@ -9,20 +9,29 @@ import (
...
+const ItemVersion uint64 = 1
+
 // Item is the queue representation of an item to publish to the queue.
 type Item struct {
 	Build    *library.Build  `json:"build"`
 	Pipeline *pipeline.Build `json:"pipeline"`
 	Repo     *library.Repo   `json:"repo"`
 	User     *library.User   `json:"user"`
+	// The 0-value is the implicit ItemVersion for queued Items that pre-date adding the field.
+	ItemVersion uint64 `json:"item_version"`
```
cognifloyd, line 25: NB: This is downgrade safe because if a json object has an unknown field, it is ignored.

wass3r (approved):

PR #293, opened 2023-05-19
revert(repo): add AllowSchedule field

jbrockopp:
Reverts go-vela/types#291 based off of https://github.com/go-vela/types/pull/291#issuecomment-1554711054

PR #294, opened 2023-05-22
enhance(repo)!: add AllowSchedule field v2

jbrockopp:
xref: https://github.com/go-vela/community/issues/538

Part of https://github.com/go-vela/community/pull/772

Based off of https://github.com/go-vela/types/pull/289

Duplicate of https://github.com/go-vela/types/pull/293

This change add an `AllowSchedule` field to the `Repo{}` object:

https://github.com/go-vela/types/blob/e3e6b59f4543fa22095da8f51a03b929211dea23/database/repo.go#L71

https://github.com/go-vela/types/blob/e3e6b59f4543fa22095da8f51a03b929211dea23/library/repo.go#L38

Similar to the other `Allow*` fields, this will enable users to control if their repo responds to the `schedule` event.

Initially, I didn't plan to add this field since the `Allow*` fields control which `git` based events to respond to.

And `schedules` will be totally controlled by Vela (i.e. not SCM driven), and won't be an event for webhooks.

That being said, I wanted to start with opening this PR to collect feedback on if we should add this or not.

You can find an example of how this field will be used within https://github.com/go-vela/server/pull/846.

> NOTE: Marking this as a breaking change since it adds a new field which will require DB migrations.


wass3rw3rk:
@jbrockopp do we need/want this?

jbrockopp:
@wass3rw3rk I'm personally in favor of not adding this since I don't think it provides value.

My line of thinking is someone can already disable the ability to run schedules for a repo by:

* ignore creating schedules for a repo
* if schedules already exist for a repo, disable all active schedules
* if schedules already exist for a repo, remove all active schedules

I'm personally in favor of encouraging those patterns instead of a repo level toggle because it impacts the server.

i.e. every active schedule in the database is scanned and processed by the server

PR #302, opened 2023-08-08
feat(queuing)!: create new types for `BuildExecutable` 

ecrupper:
Part of the effort for https://github.com/go-vela/community/pull/805.

This type will be pushed to a table and "popped" by a worker when it is ready to run a build. This eliminates all sensitive information (namely netrc) from the queue object.

### DIFF
--- a/library/build_executable_test.go
+++ b/library/build_executable_test.go
@@ -86,13 +86,13 @@ func TestLibrary_BuildExecutable_String(t *testing.T) {
 	bExecutable := testBuildExecutable()
 
 	want := fmt.Sprintf(`{
+  ID: %d,
   BuildID: %d,
   Data: %s,
-  ID: %d,
 }`,
+		bExecutable.GetID(),
 		bExecutable.GetBuildID(),
 		bExecutable.GetData(),
-		bExecutable.GetID(),
 	)
 
 	// run test


item.go
```diff
@@ -17,18 +16,16 @@ const ItemVersion uint64 = 1
```
plyr4, line 16: should we bump `ItemVersion` to 2?

ecrupper: good call, will mark this PR as breaking too

timhuynh94 (commented):
Can you change the constants files to 2023 as well?

ecrupper:
> Can you change the constants files to 2023 as well?

I'm waiting for movement on [this](https://github.com/go-vela/community/issues/747) before updating copyright years, which will probably result in a mass change to our repos file headers.

library/build_executable.go
```diff
@@ -0,0 +1,110 @@
...
+func (b *BuildExecutable) String() string {
+	return fmt.Sprintf(`{
+  BuildID: %d,
+  Data: %s,
+  ID: %d,
+}`,
+		b.GetBuildID(),
+		b.GetData(),
+		b.GetID(),
+	)
```
wass3rw3rk, line 109: super minor suggestion: keep order of fields same as they are defined in struct?

wass3rw3rk (approved):
one super minor suggestion, but lgtm

wass3rw3rk (approved):

PR #303, opened 2023-08-16
enhance(log): improve secret masking to better handle URL strings

ecrupper:
Adding `?` and `&` to masking buffer and adding a test. Extra layer of protection.

### DIFF
--- a/library/log_test.go
+++ b/library/log_test.go
@@ -54,6 +54,8 @@ func TestLibrary_Log_MaskData(t *testing.T) {
 	s3Masked := "$ echo $SECRET1\n***\n$ echo $SECRET2\n***\n"
 	s4 := "SOME_SECRET=((%.YY245***pP.><@@}}"
 	s4Masked := "SOME_SECRET=***"
+	s5 := "www.post-secrets.com?username=secret&password=extrasecret"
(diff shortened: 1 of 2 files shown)


library/log_test.go
```diff
@@ -54,6 +54,8 @@ func TestLibrary_Log_MaskData(t *testing.T) {
 	s3Masked := "$ echo $SECRET1\n***\n$ echo $SECRET2\n***\n"
 	s4 := "SOME_SECRET=((%.YY245***pP.><@@}}"
 	s4Masked := "SOME_SECRET=***"
+	s5 := "www.post-secrets.com?username=secret&password=extrasecret"
+	s5Masked := "www.post-secrets.com?username=***&password=***"
```
wass3rw3rk, line 58:
```suggestion
	s5 := "www.example.com?username=secret&password=extrasecret"
	s5Masked := "www.example.com?username=***&password=***"
```

suggestion: use reserved url intended for examples vs random url (see https://datatracker.ietf.org/doc/html/rfc2606#section-3)

timhuynh94 (approved):

wass3rw3rk (approved):

PR #304, opened 2023-08-21
enhance(executable): encrypt DB executable

ecrupper:
Ref: https://github.com/go-vela/server/pull/927#discussion_r1297772714

Rows will include sensitive information, so it is best to encrypt the data.

### DIFF
--- a/database/build_executable.go
+++ b/database/build_executable.go
@@ -86,10 +86,10 @@ func (b *BuildExecutable) Decrypt(key string) error {
 	return nil
 }
 
-// Encrypt will manipulate the existing repo hash by
+// Encrypt will manipulate the existing build executable by
 // creating a AES-256 cipher block from the encryption
-// key in order to encrypt the repo hash. Then, the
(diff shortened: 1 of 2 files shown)


database/build_executable.go
```diff
@@ -45,15 +46,63 @@ func (c *BuildExecutable) Compress(level int) error {
...
+	b.Data = decrypted
+
+	return nil
+}
+
+// Encrypt will manipulate the existing repo hash by
+// creating a AES-256 cipher block from the encryption
+// key in order to encrypt the repo hash. Then, the
+// repo hash is base64 encoded for transport across
+// network boundaries.
```
plyr4, line 93:
```suggestion
// Encrypt will manipulate the existing build executable by
// creating a AES-256 cipher block from the encryption
// key in order to encrypt the build executable. Then, the
// build executable is base64 encoded for transport across
// network boundaries.
```

wass3rw3rk (approved):

plyr4 (approved):

PR #312, opened 2023-09-07
feat(yaml/secret): adding pull tag to secrets to create a lazy secrets ability

claire1618:
Here's the issue: https://github.com/go-vela/community/issues/196

The changes here add a tag "pull" to secrets that can be set to either "always" or "on_start". 

Setting pull to on_start ensures that secrets are pulled when the underlying step starts, and setting the pull to always is the current behavior of pulling secrets at the beginning of the build.

Example:
```
secrets:
  - name: foo
    key: go-vela/docs/foo
    engine: native
    type: repo
    pull: on_start
```


pipeline/secret_test.go
```diff
@@ -536,7 +566,7 @@ func testSecrets() *SecretSlice {
 				Image:       "vault:latest",
 				Name:        "vault",
 				Number:      1,
-				Pull:        "always",
+				Pull:        "build_start",
```
ecrupper, line 569: These secret plugin blocks will still be `pull: always`, as that is referring to the vault image

yaml/secret.go
```diff
@@ -25,6 +25,7 @@ type (
 		Engine string `yaml:"engine,omitempty" json:"engine,omitempty" jsonschema:"enum=native,enum=vault,default=native,description=Name of storage backend to fetch secret from.\nReference: https://go-vela.github.io/docs/reference/yaml/secrets/#the-engine-tag"`
 		Type   string `yaml:"type,omitempty"   json:"type,omitempty" jsonschema:"enum=repo,enum=org,enum=shared,default=repo,description=Type of secret to fetch from storage backend.\nReference: https://go-vela.github.io/docs/reference/yaml/secrets/#the-type-tag"`
 		Origin Origin `yaml:"origin,omitempty" json:"origin,omitempty" jsonschema:"description=Declaration to pull secrets from non-internal secret providers.\nReference: https://go-vela.github.io/docs/reference/yaml/secrets/#the-origin-tag"`
+		Pull   string `yaml:"pull,omitempty"   json:"pull,omitempty" jsonschema:"default=always,description=Where to pull in secrets from storage backend."`
```
ecrupper, line 28:
```suggestion
		Pull   string `yaml:"pull,omitempty"   json:"pull,omitempty" jsonschema:"default=build_start,description=When to pull in secrets from storage backend."`
```

ecrupper (changes requested):

ecrupper (approved):

cognifloyd (approved):

PR #314, opened 2023-09-15
feat(repo)!: change `allow_<event>` fields to an `AllowEvents` struct + DB use integer masking

ecrupper:
Ref: https://github.com/go-vela/community/pull/828

This change lays the groundwork for enhanced event support moving forward by leveraging a bit-shifting iota type for allowed events represented in the database and a new struct for `AllowEvents` in the library.

This helps future development of events (and event actions) by not impacting the database. Plus, it allows a clean representation of data for the UI to interpret when the `AllowEvents` struct can be transformed into a nested checklist.


constants/allow_events.go
```diff
@@ -0,0 +1,17 @@
...
+// Allowed repo events.
+const (
+	AllowPushBranch = 1 << iota // 00000001 = 1
+	AllowPullOpen               // 00000010 = 2
+	AllowPullEdit               // 00000100 = 4
+	AllowPullSync               // ...
+	AllowPushTag
+	AllowDeployCreate
+	AllowCommentCreate
+	AllowCommentEdit
```
ecrupper, line 16: A note on this: we can't shuffle these at any point if this were to be implemented. So, if a `pull_request` event with the action `review_requested` is something we support later, then it must go at the bottom.

wass3rw3rk: seems like a good thing to add to comments in-line

cognifloyd: Can we introduce some kind of version bit (byte?) so that the entire thing can be invalidated in the future? For now, just include the bit. Then, the migration logic can be introduced when we bump that version, if we ever do so.

library/event_actions.go
cognifloyd, line 1: There are a lot of objects in here. Can we split this into
- `library/events/push_actions.go`
- `library/events/pull_actions.go`
- `library/events/deploy_actions.go`
- `library/events/comment_actions.go`

constants/allow_events.go
cognifloyd: Since we're using Intl64, and so far this only uses 8 bits, So, the version info could actually be at the end of the 64 bits instead of the beginning. Then, we don't need to reserve any space.

ecrupper: My hesitation with including a version is that it would require some sort of DB migration script that updates all existing repos every time we add a new event we support. Part of me wants to just go in and enumerate all possible webhook actions that Vela could theoretically subscribe to and just mark them as `not implemented` if they aren't. That way future enhancements would build off a clean order and it wouldn't change unless GitHub itself adds a new event / action worth implementing.

cognifloyd: Yes. Let's leave room for all current GitHub events, but leave a `_` for all of the unimplemented ones with a comment saying what that slot is reserved for. Maybe even leave a bit of a gap in each category to add more later.

ecrupper: Added space for future implementations for `pull_request` actions that seem potentially relevant to CI

ecrupper:
After some discussion amongst the @go-vela/admins, we are deciding to:

1. Move this to `v0.23.0`
2. Keep the DB fields of `AllowPush`, `AllowPull`, `AllowDeploy`, `AllowTag`, and `AllowComment` to preserve the data when we decide to upgrade to this new `AllowEvents` model. This will make migration scripting easier.
3. Deprecate those fields and then plan on dropping them with the release of `v0.24.0`.

constants/allow_events.go
cognifloyd: There seems to be a lot of other possible webhook events:
https://docs.github.com/en/webhooks/webhook-events-and-payloads

How did you pick which of the `pull_request` actions deserved an entry here?

library/repo.go
```diff
@@ -705,15 +736,46 @@
...
+	case constants.EventDeploy:
+		allowed = r.GetAllowEvents().GetDeployment().GetCreated()
+	}
+
+	return
+}
+
 // String implements the Stringer interface for the Repo type.
 func (r *Repo) String() string {
 	return fmt.Sprintf(`{
```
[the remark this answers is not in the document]
ecrupper: No idea why this just surfaced

KellyMerrick (changes requested):

KellyMerrick (approved):
lgtm

timhuynh94 (approved):

wass3rw3rk (approved):
let's go to the moon 🌑

PR #323, opened 2023-10-02
feat(deployment): add deployment table

claire1618:
When a repo has a lot of deployments historically, viewing them in the UI is cumbersome and slow.

The loading of deployments on the deployments tab in Vela takes a few more seconds to load than the other pages (schedules, secrets, audits...) even when there are no deployments for a repo. This is due to the way deployments are stored and retrieved.

The changes made in this pull request add a new database table so that we can store deployment information instead of having to go to the scm to retrieve them. This (plus the implementation in server) greatly improve the time it takes for the deployments page to load.


wass3rw3rk:
release and other merges caused some conflicts, so you'll have to resolve that first

wass3rw3rk:
also, we will want to add tests

KellyMerrick (changes requested):

timhuynh94 (approved):

database/deployment.go
```diff
@@ -0,0 +1,176 @@
...
+	URL         sql.NullString     `sql:"url"`
+	User        sql.NullString     `sql:"user"`
+	Commit      sql.NullString     `sql:"commit"`
+	Ref         sql.NullString     `sql:"ref"`
+	Task        sql.NullString     `sql:"task"`
+	Target      sql.NullString     `sql:"target"`
+	Description sql.NullString     `sql:"description"`
+	Payload     raw.StringSliceMap `sql:"payload"`
+	Builds      pq.StringArray     `sql:"builds" gorm:"type:varchar(50)"`
+}
```
wass3rw3rk, line 39: suggestion: we might want to add Created as we do in other tables so we can more easily purge these

wass3rw3rk (changes requested):

database/deployment.go
claire1618: added!

database/build.go
```diff
@@ -45,6 +45,7 @@ type Build struct {
 	Started       sql.NullInt64      `sql:"started"`
 	Finished      sql.NullInt64      `sql:"finished"`
 	Deploy        sql.NullString     `sql:"deploy"`
+	DeploymentID  sql.NullInt64      `sql:"deployment_id"`
```
ecrupper, line 48:
```suggestion
	DeployID  sql.NullInt64      `sql:"deployment_id"`
```
Thoughts on this for consistency with `Deploy` and `DeployPayload`?

database/build.go
```diff
@@ -170,6 +171,11 @@ func (b *Build) Nullify() *Build {
 		b.Deploy.Valid = false
 	}
 
+	// check if the Deploy field should be false
```
ecrupper, line 174:
```suggestion
	// check if the DeployID field should be false
```

database/deployment.go
```diff
@@ -0,0 +1,183 @@
...
+	d.Ref = sql.NullString{String: sanitize(d.Ref.String), Valid: d.Ref.Valid}
+	d.Task = sql.NullString{String: sanitize(d.Task.String), Valid: d.Task.Valid}
+	d.Target = sql.NullString{String: sanitize(d.Target.String), Valid: d.Target.Valid}
+	d.Description = sql.NullString{String: sanitize(d.Description.String), Valid: d.Description.Valid}
+
+	return nil
+}
+
+// DeploymentFromLibrary converts the Deployment type
+// to a library Deployment type.
```
ecrupper, line 159:
```suggestion
// DeploymentFromLibrary converts the library Deployment type
// to a database Deployment type.
```

database/deployment_test.go
```diff
@@ -0,0 +1,174 @@
...
+	want.SetTarget("production")
+	want.SetDescription("Deployment request from Vela")
+	want.SetPayload(raw.StringSliceMap{"foo": "test1"})
+	want.SetCreatedAt(1)
+	want.SetCreatedBy("octocat")
+	want.SetBuilds(nil)
+
+	got := testDeployment().ToLibrary(nil)
+	if diff := cmp.Diff(got, want); diff != "" {
+		t.Errorf("(ToLibrary: -want +got):\n%s", diff)
```
ecrupper, line 83: The order of want and got have to match in the error message as they do in the function call

database/deployment_test.go
```diff
@@ -0,0 +1,174 @@
...
+	d.SetURL("https://github.com/github/octocat/deployments/1")
+	d.SetCommit("1234")
+	d.SetRef("refs/heads/main")
+	d.SetTask("deploy:vela")
+	d.SetTarget("production")
+	d.SetDescription("Deployment request from Vela")
+	d.SetPayload(raw.StringSliceMap{"foo": "test1"})
+	d.SetCreatedAt(1)
+	d.SetCreatedBy("octocat")
+	d.SetBuilds(nil)
```
ecrupper, line 145: I'd like to see some unit testing on when this field is actually populated. Just a simple build partial or two should work as long as the ID is set.

library/build.go
```diff
@@ -158,6 +159,7 @@ func (b *Build) Environment(workspace, channel string) map[string]string {
 		envs["VELA_BUILD_TARGET"] = target
 		envs["VELA_DEPLOYMENT"] = target
 		envs["BUILD_TARGET"] = target
+		envs["VELA_DEPLOYMENT_NUMBER"] = ToString(b.GetDeploymentID())
```
ecrupper, line 162: One of the benefits of storing our own deployment record is we can create our own `Environment` function for the deployment type. Might be worth considering here. At bare minimum, this should be `VELA_DEPLOYMENT_ID`, which is arguably not worth including 😅

ecrupper (changes requested):
A few random things I noticed + thoughts on giving `library.Deployment` its own `Environment()` method? I can imagine that could be useful to some users.

library/build.go
wass3rw3rk, line 162:
```suggestion
		envs["VELA_DEPLOYMENT_ID"] = ToString(b.GetDeploymentID())
```

library/build_test.go
```diff
@@ -226,6 +228,7 @@ func TestLibrary_Build_Environment(t *testing.T) {
 				"VELA_BUILD_TITLE":         "push received from https://github.com/github/octocat",
 				"VELA_BUILD_WORKSPACE":     "TODO",
 				"VELA_DEPLOYMENT":          "production",
+				"VELA_DEPLOYMENT_NUMBER":   "0",
```
wass3rw3rk, line 231:
```suggestion
				"VELA_DEPLOYMENT_ID":   "0",
```

library/build_test.go
```diff
@@ -286,6 +289,7 @@ func TestLibrary_Build_Environment(t *testing.T) {
...
```
wass3rw3rk, line 292:
```suggestion
				"VELA_DEPLOYMENT_ID":   "0",
```

wass3rw3rk (changes requested):
one suggestion.. got a little confused with ID and Number 😵‍💫 let me know what you think - looks like Easton mentioned this also :disappear:

library/build.go
wass3rw3rk: that's fair - i like that

wass3rw3rk (approved):

plyr4 (approved):
looks great.

library/deployment.go
```diff
@@ -13,16 +13,18 @@ import (
 // swagger:model Deployment
 type Deployment struct {
 	ID          *int64              `json:"id,omitempty"`
+	Number      *int64              `json:"number,omitempty"`
 	RepoID      *int64              `json:"repo_id,omitempty"`
 	URL         *string             `json:"url,omitempty"`
-	User        *string             `json:"user,omitempty"`
```
plyr4, line 18: just curious, why'd we remove User?

ecrupper (approved):

library/deployment.go
claire1618: because I added a CreatedBy field instead

KellyMerrick (approved):
lgtm

PR #326, opened 2023-10-06
fix(rulesets): surface error if regex compilation fails

ecrupper:
Will lead to minor downstream changes in server and worker

Ref: https://github.com/go-vela/community/issues/435

Crashing the server when a user supplies bad regex is an overreaction 😅 

Also did some rewriting of the `Match` code path to limit some complexity and duplication. Tests were very thorough, so the rewrite should have zero impact.


ecrupper:
Closing this and reopening as GitHub is failing to catch recent commits....

PR #327, opened 2023-10-09
fix(rulesets): surface error when regexp compilation fails + some match code refactoring

ecrupper:
Previously, whenever users would supply bad regex, the server would crash with a panic due to the `MustCompile` call. 

Ref: https://github.com/go-vela/community/issues/435

I also changed up some of the match code in order to address some of the linter feedback (complexity + function length). No behavior changes.


pipeline/container.go
```diff
@@ -189,31 +194,46 @@ func (c *Container) Execute(r *RuleData) bool {
...
+	match, err := c.Ruleset.Match(r)
+	if err != nil {
+		return false, err
+	}
+
 	// check if you need to skip a status failure ruleset
 	if strings.EqualFold(status, constants.StatusSuccess) &&
 		!(c.Ruleset.If.NoStatus() && c.Ruleset.Unless.NoStatus()) &&
-		!(c.Ruleset.If.Empty() && c.Ruleset.Unless.Empty()) && c.Ruleset.Match(r) {
+		!(c.Ruleset.If.Empty() && c.Ruleset.Unless.Empty()) && match {
```
wass3rw3rk, line 222: any concerns that .Match will be called for sure now whereas in the previous version other checks had to pass before it would _maybe_ call .Match?

wass3rw3rk: sorry, maybe i should have phrased it differently. i was trying to say that .Match is a bit more intensive to call than the other checks before it. now we are calling the more intensive things for sure. the difference might be negligible, however. just pointing it out :D

pipeline/ruleset.go
```diff
@@ -122,170 +125,170 @@ func (r *Rules) Empty() bool {
...
-				return true
+			matches, err := matches(r, from, matcher, p, constants.OperatorAnd)
+			if err != nil {
+				return false, err
+			}
+
+			// return true if any ruletype matches the ruledata
+			if matches {
+				return true, nil
 			}
```
wass3rw3rk, line 162: could do `return matches(r, from, matcher, p, constants.OperatorAnd)`

pipeline/ruleset.go
```diff
@@ -122,170 +125,170 @@ func (r *Rules) Empty() bool {
...
-			r.Target.MatchOr(from.Target, matcher) {
-			return true
+		if matches {
+			return true, nil
 		}
 
 		// return false if no match is found
-		return false
+		return false, nil
 	}
```
wass3rw3rk, line 184: could also be `return matches(r, from, matcher, "", constants.OperatorOr)`

pipeline/ruleset.go
```diff
@@ -122,170 +125,170 @@ func (r *Rules) Empty() bool {
...
-					r.Event.MatchOr(from.Event, matcher) ||
-					r.Path.MatchOr(p, matcher) ||
-					r.Repo.MatchOr(from.Repo, matcher) ||
-					status ||
-					r.Tag.MatchOr(from.Tag, matcher) ||
-					r.Target.MatchOr(from.Target, matcher) {
-					return true
+				if matches {
+					return true, nil
 				}
```
wass3rw3rk, line 145: reduce to `return matches(r, from, matcher, p, constants.OperatorOr)`?

pipeline/ruleset.go
```diff
@@ -122,170 +125,170 @@ func (r *Rules) Empty() bool {
...
-		return true
+	// return true if any ruletype matches the ruledata
+	if matches {
+		return true, nil
 	}
 
 	// return false if no match is found
-	return false
+	return false, nil
 }
```
wass3rw3rk, line 198: `return matches(r, from, matcher, "", constants.OperatorAnd)` maybe?

wass3rw3rk (commented):
not sure if you're being explicit intentionally, otherwise we can get rid of some `if` statements

pipeline/ruleset.go
ecrupper: We want to continue looping through the paths if there's no match, so returning `false, nil` is not an option here unfortunately...

pipeline/ruleset.go
ecrupper: Done 👍

pipeline/ruleset.go
ecrupper: Same idea with the looping situation as above ^^

pipeline/ruleset.go
ecrupper: Done 👍

pipeline/ruleset.go
wass3rw3rk: hmm, maybe we need to adjust tests since i didn't see any test fails with the change?!

pipeline/build.go
```diff
@@ -33,34 +33,54 @@ type Build struct {
 // function will remove the steps that have a ruleset that do not
 // match the provided ruledata. If both stages and steps are
 // provided, then an empty pipeline is returned.
-func (b *Build) Purge(r *RuleData) *Build {
+func (b *Build) Purge(r *RuleData) (*Build, error) {
```
plyr4, line 36: from first glance i didnt see a PR to update the method call in server
https://github.com/go-vela/server/blob/main/compiler/native/expand.go#L92

i wonder if returning `err` has an underlying complication with that `continue`? because we wouldve normally just returned an empty pipeline and moved on but now we have a couple different errors that might occur

plyr4: im not sure though, ill have to test that myself with some bogus pipelines

plyr4: tested it out, it bubbles up to an audit failure very nicely. nice work

plyr4 (approved):

claire1618 (approved):

PR #328, opened 2023-10-10
feat(repo settings): approve build mechanism for `pull_request` events

ecrupper:
### DESCRIPTION
Inspired by [this behavior](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) that GH Actions is able to do.

This functionality would protect our user base from untrusted sources executing pipelines without approval. It is also much more tenable to provide this capability rather than show warnings like [this](https://github.com/go-vela/ui/blob/fb659119e158e120d50e4d663c7b86eabe954ff4/src/elm/Pages/RepoSettings.elm#L453-L465). 

This security enhancement was made possible by the temporary storage of compiled build objects ([main PR for that](https://github.com/go-vela/server/pull/927)). Using this, we can put builds on the ice box while they await approval.

The below diagrams illustrate this process (note the two DB icons are the same database, just duplicated for readability):

### BEFORE
Screenshot 2023-10-10 at 2 24 22 PM

### AFTER
Screenshot 2023-10-10 at 2 25 38 PM

### MISC
As an aside, this will close all these open PRs for trying to differentiate PR events from forks:
Closes https://github.com/go-vela/types/pull/271
Closes https://github.com/go-vela/server/pull/727
Closes https://github.com/go-vela/ui/pull/593


constants/repo.go
```diff
@@ -16,3 +16,21 @@ const (
...
+	// all builds triggered from a forked PR.
+	ApproveAlways = "always"
+
+	// ApproveNoWrite defines the CI strategy of having a repo administrator approve
+	// all builds triggered from a forked PR where the author does not have write access.
+	ApproveNoWrite = "no-write"
+
+	// ApproveOnce defines the CI strategy of having a repo administrator approve
+	// all builds triggered from an outside contributor if this is their first time contributing.
+	ApproveOnce = "first-time"
```
ecrupper, line 32: A note on this: I am adding this constant as more of a To-Do. I don't have actual functionality for this coded up, but would create a follow-up story to implement this at a later point.

library/repo.go
```diff
@@ -393,6 +395,19 @@ func (r *Repo) GetPreviousName() string {
 	return *r.PreviousName
 }
 
+// GetApproveForkBuild returns the ApproveForkBuild field.
+//
+// When the provided Repo type is nil, or the field within
+//  the type is nil, it returns the zero value for the field.
```
cognifloyd, line 401:
```suggestion
// the type is nil, it returns the zero value for the field.
```

cognifloyd (approved):

cognifloyd (dismissed):

timhuynh94 (dismissed):

ecrupper:
Slotting for `v0.23`

wass3rw3rk (commented):
a couple of comments

## consider other forms of approval?
i wonder if we want to decouple this from fork builds more (at the least the bit attached to the repo object) and maybe push that aspect to the value(s) you can provide. reason is to avoid potential issues later when we want to (maybe?) support other types of approval? for example, maybe we want to support approval for `tag` builds as these are often associated with prod deploys.

on the other hand, if you wanted this functionality, i _think_ you can kind of get there with deployments today by utilizing https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#required-reviewers . that is, send a deployment request via Vela, and then it hangs out in GitHub until someone approves. once approved it sends off the webhook for the deployment to Vela. i haven't tested this, tbh.

some folks also get more creative to get to similar functionality, like https://trstringer.com/github-actions-manual-approval/ for github actions, for example.

## track approver
i think there would be additional interest to track who approved a build (and maybe when?) from audit perspective. should we attach to build?

any thoughts on those?

ecrupper:
@wass3rw3rk Included auditing and put the fork terminology in the constants rather than the field name with latest commit. Let me know what you think!

cognifloyd (approved):

timhuynh94 (dismissed):

webhook.go
```diff
@@ -23,6 +23,7 @@ type Webhook struct {
 	Hook     *library.Hook
 	Repo     *library.Repo
 	Build    *library.Build
+	PRFork   bool
```
wass3rw3rk, line 26: this doesn't seem to be used?

wass3rw3rk: nvm

wass3rw3rk (approved):

wass3rw3rk (approved):

timhuynh94 (approved):

PR #339, opened 2023-12-19
feat(dashboard): library and database types

ecrupper:
Executing on a concept where users can create pages wherein they can look at a collection of repositories at a high level. Below is a simple mock up of the eventual UI:

dashboard_mock

This back-end struct design accounts for a few key features: share-ability, admin access, and customizability.

**Share-ability**: using a uniquely generated UUID that will serve as the `link` to the dashboard along with user-specific `user.Dashboards` we will be able to allow open navigation to any dashboard via a link + easy navigation through the UI via "My Dashboards", much like many group-board applications such as Draw IO, Miro, etc.

**Admin Access**: since the dashboards will be public and discoverable, it's important to limit update/delete access. This can be done via the `Dashboard.Admins` field, which holds a slice of user_ids capable of editing the dashboard. There is also auditing in place in the form of `created_at`, `created_by`, `updated_at`, `updated_by`.

**Customizability**: choosing to store the dashboard repo collection as a JSON field allows users to customize their "subscription" to that repo, using the `Events` and `Branches` field. While it would have been possible to reference repositories in a slice field of repo_ids, that way makes customization to this degree much more challenging.

The server-side implementation is this can be found, in draft form, [here](https://github.com/go-vela/server/pull/1028)


ecrupper:
Closing with migration of types -> server: https://github.com/go-vela/server/pull/1028

PR #340, opened 2023-12-29
feat: adding a delete_event

claire1618:
This code enables the delete webhook in Vela, so users can select to trigger a build when a branch or tag are deleted.


database/repo.go
```diff
@@ -66,6 +66,7 @@ type Repo struct {
 	AllowDeploy  sql.NullBool   `sql:"allow_deploy"`
 	AllowTag     sql.NullBool   `sql:"allow_tag"`
 	AllowComment sql.NullBool   `sql:"allow_comment"`
+	AllowDelete  sql.NullBool   `sql:"allow_delete"`
```
ecrupper, line 69: No need for this field with the `AllowEvents` struct.

item_test.go
```diff
@@ -59,6 +59,7 @@ func TestTypes_ToItem(t *testing.T) {
 		AllowPush:   &booL,
 		AllowDeploy: &booL,
 		AllowTag:    &booL,
+		AllowDelete: &booL,
```
ecrupper, line 62: Shouldn't need this field for library.Repo

library/repo.go
```diff
@@ -47,6 +48,7 @@ func (r *Repo) Environment() map[string]string {
 	return map[string]string{
 		"VELA_REPO_ACTIVE":        ToString(r.GetActive()),
 		"VELA_REPO_ALLOW_COMMENT": ToString(r.GetAllowComment()),
+		"VELA_REPO_ALLOW_DELETE":  ToString(r.GetAllowDelete()),
```
ecrupper, line 51: Won't need these environment values

library/repo.go
```diff
@@ -35,6 +35,7 @@ type Repo struct {
 	AllowDeploy  *bool     `json:"allow_deploy,omitempty"`
 	AllowTag     *bool     `json:"allow_tag,omitempty"`
 	AllowComment *bool     `json:"allow_comment,omitempty"`
+	AllowDelete  *bool     `json:"allow_delete,omitempty"`
```
ecrupper, line 38: `AllowEvents` struct should have this info

library/repo.go
```diff
@@ -765,6 +794,8 @@ func (r *Repo) SetApproveBuild(v string) {
 }
 
 // EventAllowed determines whether or not an event is allowed based on the repository settings.
+//
+//nolint:nakedret // not necessary to watch for here
```
ecrupper, line 798: Let's follow the linter here and no longer do a naked return for this

library/secret.go
```diff
@@ -79,6 +79,8 @@ func (s *Secret) Match(from *pipeline.Container) bool {
 		eACL = checkEvent(events, constants.EventDeploy)
 	case constants.EventComment:
 		eACL = checkEvent(events, constants.EventComment)
+	case constants.EventDelete:
```
ecrupper, line 82: I think we should implement secrets allowed events the same way we do repo allowed events

ecrupper (changes requested):
There shouldn't be a new field in `library.Repo` or `database.Repo` — only in `library.Events` should there be a new action set for delete events.

library/secret.go
claire1618: Can do, let me know if the changes I made are what you imagined

wass3rw3rk (commented):
lgtm, do we want to add `delete` here as well: https://github.com/go-vela/types/blob/26e54c869418eb841c5cc1d31b3364459378c9b6/yaml/ruleset.go#L147-L163, so when you specify just `delete` it will add both actions for you? not saying we have to, just asking to get thoughts.

KellyMerrick (approved):
lgtm

ecrupper (approved):

PR #341, opened 2024-01-03
enhance(secret): use allow event system

ecrupper:
It will be helpful to share the same structure for allowed events for both repos and secrets.

### DIFF
--- a/library/actions/schedule.go
+++ b/library/actions/schedule.go
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package actions
+
+import "github.com/go-vela/types/constants"
+
+// Schedule is the library representation of the various actions associated
+// with the deploy event webhook from the SCM.
+type Schedule struct {
+	Run *bool `json:"created"`
+}
+
+// FromMask returns the Schedule type resulting from the provided integer mask.
+func (a *Schedule) FromMask(mask int64) *Schedule {
+	a.SetRun(mask&constants.AllowSchedule > 0)
+
+	return a
+}
+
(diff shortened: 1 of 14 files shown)


library/actions/schedule.go
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package actions
+
+import "github.com/go-vela/types/constants"
+
+// Schedule is the library representation of the various actions associated
+// with the deploy event webhook from the SCM.
+type Schedule struct {
+	Run *bool `json:"created"`
```
wass3rw3rk, line 10:
```suggestion
	Run *bool `json:"run"`
```

library/actions/schedule.go
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package actions
+
+import "github.com/go-vela/types/constants"
+
+// Schedule is the library representation of the various actions associated
+// with the deploy event webhook from the SCM.
```
wass3rw3rk, line 8:
```suggestion
// with the schedule event webhook from the SCM.
```

wass3rw3rk (changes requested):
first pass..

also, i didn't verify, but is checkEvents() actually unused as the linter claims? we can remove, if so.

ecrupper:
> first pass..
> 
> also, i didn't verify, but is checkEvents() actually unused as the linter claims? we can remove, if so.

The linter really does have its moments

wass3rw3rk (dismissed):

wass3rw3rk (approved):

PR #344, opened 2024-01-16
enhance(ci): include PR title validation workflow

ecrupper:
Curious what the thoughts are on something like this. I know we use a [script](https://github.com/go-vela/community/blob/main/releases/scripts/relgen.sh) to generate release notes that relies on this pattern.


wass3r:
Yes, please. Extra points for not using a third party action 👍 make it required check for all repos.

wass3r (approved):

claire1618 (approved):

PR #348, opened 2024-01-19
feat: remodling EventDelete to resemble EventTag

claire1618:
Every time a tag or branch are deleted in GitHub, both a delete and push webhook are sent out. Previously I used the delete webhook and had EventDelete modeled off of EventPush, EventPR, EventDeploy, and EventComment in how they are their own event not related to the others. However, the push webhook that is sent out with the delete webhook contains more relevant information that can solve some of the issues that were found with only having the delete webhook data (such as missing a commit and not being able to properly set the BUILD_TAG environment variable).

Therefore, EventDelete has been lumped in with push events, in the same way that EventTag has been.


library/actions/push.go
```diff
@@ -8,14 +8,18 @@ import "github.com/go-vela/types/constants"
 // Push is the library representation of the various actions associated
 // with the push event webhook from the SCM.
 type Push struct {
-	Branch *bool `json:"branch"`
-	Tag    *bool `json:"tag"`
+	Branch       *bool `json:"branch"`
+	Tag          *bool `json:"tag"`
+	DeleteBranch *bool `json:"delete:branch"`
+	DeleteTag    *bool `json:"delete:tag"`
```
wass3rw3rk, line 14: i didn't check closely, but i don't think we use `:` in any other JSON keys. I'd propose either `delete_*` or `*_delete`

library/actions/push.go
```diff
@@ -57,6 +69,28 @@ func (a *Push) GetTag() bool {
 	return *a.Tag
 }
 
+// GetTag returns the DeleteBranch field from the provided Push. If the object is nil,
```
wass3rw3rk, line 72:
```suggestion
// GetDeleteBranch returns the DeleteBranch field from the provided Push. If the object is nil,
```

library/actions/push.go
```diff
@@ -57,6 +69,28 @@ func (a *Push) GetTag() bool {
...
+func (a *Push) GetDeleteBranch() bool {
+	// return zero value if Push type or Tag field is nil
+	if a == nil || a.DeleteBranch == nil {
+		return false
+	}
+
+	return *a.DeleteBranch
+}
+
+// GetTag returns the DeleteTag field from the provided Push. If the object is nil,
```
wass3rw3rk, line 83:
```suggestion
// GetDeleteTag returns the DeleteTag field from the provided Push. If the object is nil,
```

wass3rw3rk (changes requested):

library/actions/push.go
claire1618: yeah I looked around and it is main `delete_*` format so I updated that.

library/actions/push.go
```diff
@@ -57,6 +69,28 @@ func (a *Push) GetTag() bool {
 	return *a.Tag
 }
 
+// GetDeleteBranch returns the DeleteBranch field from the provided Push. If the object is nil,
+// or the field within the object is nil, it returns the zero value instead.
+func (a *Push) GetDeleteBranch() bool {
+	// return zero value if Push type or Tag field is nil
```
wass3rw3rk, line 75:
```suggestion
	// return zero value if Push type or DeleteBranch field is nil
```

library/actions/push.go
```diff
@@ -57,6 +69,28 @@ func (a *Push) GetTag() bool {
...
+		return false
+	}
+
+	return *a.DeleteBranch
+}
+
+// GetDeleteTag returns the DeleteTag field from the provided Push. If the object is nil,
+// or the field within the object is nil, it returns the zero value instead.
+func (a *Push) GetDeleteTag() bool {
+	// return zero value if Push type or Tag field is nil
```
wass3rw3rk, line 86:
```suggestion
	// return zero value if Push type or DeleteTag field is nil
```

wass3rw3rk (changes requested):
sorry, i missed these in the first go

library/build.go
```diff
@@ -202,6 +202,16 @@ func (b *Build) Environment(workspace, channel string) map[string]string {
...
 	}
 
+	// check if the Build event is delete:tag
+	if strings.EqualFold(b.GetEvent(), constants.EventDelete) && strings.EqualFold(b.GetEventAction(), constants.ActionTag) {
+		// capture the tag reference
+		tag := b.GetBranch()
+
+		// add the tag reference to the list
+		envs["BUILD_TAG"] = tag
+		envs["VELA_BUILD_TAG"] = tag
```
wass3rw3rk, line 212: i think we should add more commentary explaining why we are doing a .GetBranch for populating build tag.

wass3rw3rk (changes requested):
i think it's looking good; just one more request to provide more detail. see within.

wass3rw3rk (approved):

PR #355, opened 2024-02-05
fix(database): add limit and eviction policy for deployment builds

ecrupper:
Adding a limit + an eviction policy so we prioritize most recent builds for a given deployment.


claire1618 (approved):

KellyMerrick (approved):
lgtm

PR #365, opened 2024-03-19
fix(deployments): check for deployment created event action

plyr4:
adding support for `event: [deployment:created]`
needed by: https://github.com/go-vela/server/pull/1087

reference: https://go-vela.github.io/docs/usage/repo_settings/#webhook-events


library/events.go
```diff
@@ -66,6 +66,8 @@ func (e *Events) Allowed(event, action string) bool {
 		allowed = e.GetComment().GetEdited()
 	case constants.EventDeploy:
 		allowed = e.GetDeployment().GetCreated()
```
ecrupper, line 68:
```suggestion

```

We can ditch this case if we're setting the `EventAction` in `processDeploymentEvent` in the server

ecrupper (approved):

wass3r (dismissed):

ecrupper (approved):

PR #367, opened 2024-03-21
feat: allow for step status reporting to SCM as its own context

ecrupper:
Ref: https://github.com/go-vela/community/issues/576

Also fixes bug noted in https://github.com/go-vela/community/issues/958 by setting the ground work to _not_ report on the `pull_request` context when the action type is non-standard (opened, reopened, synchronize). Solving bugs with features!

This also lays some groundwork for potential usage of the checks API whenever Vela implements GitHub App integration

Server PR, in draft form, which may be more of interest: https://github.com/go-vela/server/pull/1090

### DIFF
--- a/constants/limit.go
+++ b/constants/limit.go
@@ -33,4 +33,7 @@ const (
 
 	// DeployBuildsMaxSize defines the maximum size in characters for deployment builds.
 	DeployBuildsMaxSize = 500
+
+	// ReportStepStatusLimit defines the maximum number of steps in a pipeline that may report their status to the SCM.
+	ReportStepStatusLimit = 10
 )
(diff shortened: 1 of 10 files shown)


constants/limit.go
```diff
@@ -33,4 +33,7 @@ const (
 
 	// DeployBuildsMaxSize defines the maximum size in characters for deployment builds.
 	DeployBuildsMaxSize = 500
+
+	// ReportStepStatusLimit defines the maximum number of steps in a pipeline that may report their status to the SCM.
+	ReportStepStatusLimit = 10
```
wass3rw3rk, line 38: 👍🏼

wass3rw3rk (approved):
do we need to account for people setting report_as to the same value? i can see that happen when folks use templates in particular maybe?

ecrupper:
> do we need to account for people setting report_as to the same value? i can see that happen when folks use templates in particular maybe?

Yeah I think that's a good point. Might be worth adding some validation in the compiler.

wass3r (dismissed):

PR #370, opened 2024-04-05
fix: allow events yaml unmarshal

plyr4:
lets `allow_events` be unmarshalled as a string slice in yaml, primarily for the cli to read add/update input from a file.

kudos @wsan3 for spotting the bug


library/repo.go
wass3rw3rk, line 1: this should probably be added in https://github.com/go-vela/server/blob/main/api/types/repo.go as well now?

wass3rw3rk (approved):

library/repo.go
plyr4: yes!

KellyMerrick (approved):
lgtm

PR #380, opened 2024-06-13
fix(container): skip "label" rule evaluation in container execution

wsan3:
## Description
Add `label` as another rule to skip when matching during container execution, as worker lacks necessary rule data.

The compiler already determines whether a container will run based on that rule and others.


ecrupper:
Can you add a test case to `container_test.go` for this?

KellyMerrick (approved):
lgtm

PR #385, opened 2024-07-14
fix(swagger): remove old model annotations

wass3r:
the modified model annotations have moved to [go-vela/server/api/types](https://github.com/go-vela/server/blob/main/api/types) where a duplicate of these annotations can be found. this change is needed for accurate swagger doc generation. see
https://github.com/go-vela/server/pull/1159 .

i've replaced the swagger documentation with godoc annotation that marks these structs as deprecated, so we should see warnings anywhere these are used yet.

### DIFF
--- a/library/build.go
+++ b/library/build.go
@@ -12,6 +12,8 @@ import (
 )
 
 // Build is the library representation of a build for a pipeline.
+//
+// Deprecated: use Build from github.com/go-vela/server/api/types instead.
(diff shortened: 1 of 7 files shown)


plyr4 (approved):

KellyMerrick (approved):
lgtm

timhuynh94 (approved):

PR #386, opened 2024-08-13
fix(yaml): remediate insecure/archived buildkite lib in favor of go-yaml v3

ecrupper:
Use `yaml.Node` to create our own `MapSlice` type that we were relying on with the buildkite/go-yamlV2 implementation. However, now anchors will work with stages, as is made obvious by the passing tests that came from https://github.com/go-vela/types/pull/12.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,11 +4,11 @@ go 1.21.9
 
 require (
 	github.com/adhocore/gronx v1.8.1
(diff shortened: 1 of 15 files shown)


timhuynh94 (approved):
LGTM

KellyMerrick (approved):
lgtm

PR #391, opened 2024-09-24
fix(yaml): support collapsing of multiple anchor keys in step

ecrupper:
https://github.com/go-vela/types/pull/386 introduced a regression(?) where the following step configurations were no longer valid:

```yaml
  - name: alpha
    <<: *alpine-image
    <<: *event-push
    commands:
      - echo alpha
```

`go-yaml` asserts that the proper usage of multiple anchors in a single map is to list them:

```yaml
  - name: beta
    <<: [ *alpine-image, *event-push ]
    commands:
      - echo beta
```

Given how many current users leverage the first configuration, this PR aims to support that format even with the upgrade to go-yaml v3.

### DIFF
--- a/yaml/service.go
+++ b/yaml/service.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"strings"
 
+	"gopkg.in/yaml.v3"
+
 	"github.com/go-vela/types/constants"
 	"github.com/go-vela/types/pipeline"
 	"github.com/go-vela/types/raw"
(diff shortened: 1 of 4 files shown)


timhuynh94 (approved):

yaml/service.go
```diff
@@ -58,21 +60,78 @@ func (s *ServiceSlice) ToPipeline() *pipeline.ContainerSlice {
...
+		newContent := new([]*yaml.Node)  // new service content
+		anchorSequence := new(yaml.Node) // final type that is appended to service contents
+
+		// iterate through map contents (key, value)
+		for i := 0; i < len(tmpService.Content); i += 2 {
+			key := tmpService.Content[i]
+			value := tmpService.Content[i+1]
+
+			// check if key is an anchor reference
+			if strings.EqualFold(key.Value, "<<") {
```
wass3rw3rk, line 96: EqualFold() seems unnecessary here (and in the other location).

wass3rw3rk (commented):
looks like we essentially have same code in two places? should we create a helper function?

yaml/service.go
plyr4: if the equalfold is necessary you could use a switch case comparison for some extra speed 
```
switch key.Value {
case "<<":
```

yaml/service.go
```diff
@@ -58,21 +60,78 @@ func (s *ServiceSlice) ToPipeline() *pipeline.ContainerSlice {
...
+
+		// initialize anchor sets
+		anchorList := new([]*yaml.Node)  // collect multiple anchor references
+		newContent := new([]*yaml.Node)  // new service content
+		anchorSequence := new(yaml.Node) // final type that is appended to service contents
+
+		// iterate through map contents (key, value)
+		for i := 0; i < len(tmpService.Content); i += 2 {
+			key := tmpService.Content[i]
+			value := tmpService.Content[i+1]
```
plyr4, line 93: im assuming this cant panic? i couldnt get it to be nil while testing

yaml/service.go
wass3rw3rk: i think it's a simple `key.Value == "<<"`, no? there are no casing concerns.

yaml/service.go
wass3rw3rk: in the same spirit, the length is guaranteed to be an even number? (i think so?) :D

yaml/service.go
plyr4: right, switch is faster tho (jump tables) but its negligible (with over 1 billion iterations it measures like 3% faster)

wass3rw3rk:
changing this to draft for now until we revisit

PR #392, opened 2024-09-26
fix(revert): use buildkite as yaml parser

ecrupper:
Due to the breaking changes in `v3` and the accelerated timeline of getting `v0.25.0` out the door, our YAML library change will occur in `v0.26`

Additional context: https://github.com/go-vela/types/pull/391


plyr4 (approved):
we'll kill you next time..... buildkite

wass3r (approved):

